PR #89, opened 2026-04-15
[Pages] Fix Web API pagination to use Prefer: odata.maxpagesize

priyanshu92:
Replace $top-based pagination with Prefer: odata.maxpagesize=N header across Web API integration skill. $top caps the total result set and prevents @odata.nextLink from being returned, breaking pagination.

- Update fetchAllPages helper to accept pageSize param and send Prefer: odata.maxpagesize header on every request
- Update listProducts and listOrderLines examples to use Prefer header instead of $top for page size control
- Update Pagination Strategy docs to explain correct approach
- Update webapi-integration agent rule 4 and completion checklist
- Bump plugin version to 1.2.1

### DIFF
--- a/plugins/power-pages/.claude-plugin/plugin.json
+++ b/plugins/power-pages/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "power-pages",
-  "version": "1.2.0",
+  "version": "1.2.1",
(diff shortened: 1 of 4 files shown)


neerajnandwana-msft (approved):
Changes looks good to me.

PR #103, opened 2026-04-17
[Canvas Apps] Add skill to migrate apps with SP list data sources to Dataverse Tables

rsantos00:
Add a new skill to migrate Apps with Sharepoint list datasources to Dataverse tables.

### DIFF
--- a/plugins/canvas-apps/skills/migrate-to-dataverse/SKILL.md
+++ b/plugins/canvas-apps/skills/migrate-to-dataverse/SKILL.md
@@ -1,0 +1,111 @@
+---
+name: migrate-to-dataverse
+version: 1.0.0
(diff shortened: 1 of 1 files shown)


rsantos00:
@microsoft-github-policy-service agree

rsantos00:
@microsoft-github-policy-service agree company="Microsoft"

plugins/canvas-apps/skills/migrate-to-dataverse/SKILL.md
```diff
@@ -0,0 +1,111 @@
+---
+name: migrate-to-dataverse
```
lesaltzm, line 2: Does this skill need to specifically be migrating to dataverse, or could it be genericized to any table-replacement operations?

rsantos00: I think the Power Fx syntax might be tricky if we have think of SQL Views/Tables and the 95% of migration cases are SP-> Dataverse. Based on customers we should start with this one and if we get customer asking for SQL or any datasources, we can enhance with more targeted skills.

PR #114, opened 2026-04-21
Add OpenCode installer support

ulrich-zogo:
## Summary
- add OpenCode support to the installer by generating namespaced filesystem-based skill and agent wrappers alongside the existing Claude Code and GitHub Copilot marketplace flow
- update README and plugin docs, plus version-check guidance, so OpenCode users get the right install, update, and command names
- normalize generated OpenCode agent frontmatter and ignore project-local .opencode/ files so local state does not break the CLI or leak into git

## Testing
- node --check scripts/install.js
- node --test plugins/power-pages/scripts/tests/check-version.test.js
- opencode --help

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ ENV/
 
 # Local settings
 .claude/settings.local.json
+.opencode/
 
 # OS
 .DS_Store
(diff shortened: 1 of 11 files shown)


ulrich-zogo:
@microsoft-github-policy-service agree

PR #117, opened 2026-04-22
[Pages] Use Power Pages icon in create-site preview loader

priyanshu92:
- Replace generic 3-slab CSS logo in the create-site scaffold loading screen with the official Power Pages PNG icon.
- Applied across all four framework scaffolds: react, vue, astro, angular
- Dropped the per-slab DOM children and their keyframes; kept a gentle coreFloat breathing animation so the loader still feels alive


priyanshu92:
Good call — addressed in 01e9e8e. Moved the PNG to `public/power-pages-icon.png` in each scaffold (react, vue, astro, angular) and the CSS now references it via `url('/power-pages-icon.png')`. Removes ~30 KB of inlined base64 per file and makes future edits a simple asset swap.

ashishchoudhary001 (approved):

PR #142, opened 2026-05-06
feat(telemetry) 1/3: shared library — dispatcher, consent, sync, builders     

amitjoshi438:
Adds the shared 1DS telemetry library used by adopting plugins. Default-on, env-var opt-out only (`POWER_PLATFORM_SKILLS_TELEMETRY=0`). No plugin adopts it in this PR — PR 2 does that.
                             
  **Cannot emit on merge.** `shared/telemetry/ikey.json` ships with `"disabled": true`. The dispatcher exits before network/local-log paths run. A separate PR flips the flag after the
  tenant-side Kusto stream is provisioned.                                                                                                                                                      
                                                                 
  ## Review focus                                                                                                                                                                               
                                                         
  | Priority | Files | What to verify |                                                                                                                                                         
  |---|---|---|                                          
  | 🔴 Read carefully | `lib/events.js` | Field allowlist matches design doc §4. `FIELD_TYPES` is the privacy boundary — fields not listed cannot reach the wire. |
  | 🔴 | `lib/emit-dispatcher.js` | Envelope has exactly 5 keys (`ver`, `name`, `time`, `iKey`, `data`); body ends with `\n`; kill switch (L53) + env opt-out (L27) gate emission. |
  | 🔴 | `lib/emit-spawn.js` | Restricted child env (no ambient secrets); detached + unref'd so parent never waits. |
  | 🟡 Scan | `lib/{pac-auth,agent-info}.js` | Only GUIDs (org/tenant) or version strings ever read. Caches per-process, never throws. |
  | 🟡 | `lib/{correlation,session,scrubber,local-log,prompt-detector,emit-from-prompt,with-telemetry}.js` | Stateless utilities + orchestrators. |
  | ⚪ Skim | `tests/*.test.js` (109 tests) | Hermetic — no real network, no real PAC. `FAKE_HTTPS` probe asserts wire shape. |
  | ⚪ | `sync-to-plugin.js` + design docs + repo `AGENTS.md` | Reference. |

  ## Not in scope

  - No plugin adopts this library here (see PR 2).
  - No first-run prompt, no consent file. Single opt-out is `POWER_PLATFORM_SKILLS_TELEMETRY=0`.
  - Zero npm dependencies; pure Node stdlib.


  Stacked

  This is PR 1 of 3. PR 2 wires power-pages to it; PR 3 adds the slash-command hook.


shared/telemetry/ikey.json
```diff
@@ -0,0 +1,6 @@
+{
+  "ikey": "ffdb4c99ca3a4ad5b8e9ffb08bf7da0d-65357ff3-efcd-47fc-b2fd-ad95a52373f4-7402",
```
priyanshu92: Let's use the full form here. Also, we'll need to have separate keys for each cloud. Reference https://github.com/microsoft/powerplatform-vscode/blob/4d029752a83d34384643e026681f89664514f7f2/src/common/OneDSLoggerTelemetry/telemetryConstants.ts#L61-L77

shared/telemetry/lib/emit-dispatcher.js
```diff
@@ -0,0 +1,147 @@
...
+// PRs land while the tenant-side annotation + Kusto table are still being
+// provisioned. Flip to false in a single PR when ready.
+//
+// Test seam: POWER_PLATFORM_SKILLS_BYPASS_KILL_SWITCH=1 forces the gate
+// open so existing dispatcher tests can exercise the emission paths
+// without round-tripping through the shared ikey.json.
+function isDisabledByConfig() {
+  if (process.env.POWER_PLATFORM_SKILLS_BYPASS_KILL_SWITCH === "1") {
+    return false;
+  }
```
priyanshu92: Can we remove this? We should have a way to mock this for tests.

shared/telemetry/lib/emit-dispatcher.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+function buildEnvelope(event) {
+  return {
+    ver: "4.0",
+    name: event.name,
+    time: new Date().toISOString(),
+    iKey: "o:" + IKEY.split("-")[0],
+    data: event.data || {},
```
priyanshu92: How are we enforcing `FIELD_TYPES` here? This will allow logging anything, right?

shared/telemetry/lib/with-telemetry.js
```diff
@@ -0,0 +1,111 @@
...
+  let caught;
+  try {
+    return await asyncFn();
+  } catch (err) {
+    outcome = "failure";
+    errorClass = err && err.constructor ? err.constructor.name : "Error";
+    errorDescription =
+      err && err.message
+        ? String(err.message).slice(0, ERROR_DESCRIPTION_MAX_LEN)
+        : "";
```
priyanshu92: The `err.message` can contain system paths which will be PII. Let's scrub that out.

AI Suggestion:

>Suggested fix: Either: (a) restrict errorDescription to err.code/err.errno/err.name (non-PII metadata) and never the free-form message; or (b) implement a real scrubber (at minimum: strip /Users/<name>, /home/<name>, C:\Users\<name>, file:// URLs, https?://<creds>@ auth URLs, IPs, GUIDs, and tokens), wire it into with-telemetry.js:84-87, and add tests that assert each of those patterns is redacted.

amitjoshi438: errorDescription is now err.code only. err.message is never read. New test asserts a PII-bearing message doesn't appear in the payload.

shared/telemetry/ikey.json
amitjoshi438: Updated to use the full form, will add multi cloud in follow up PR after testing the e2e flow on Test cluster.

shared/telemetry/lib/emit-dispatcher.js
amitjoshi438: There was a filter at event builder entry point, added here also now as a second check.

PR #149, opened 2026-05-11
force-link-environment skill + setup-pipeline Pattern 15 auto-fix

tyaginidhi:
## Summary

- New `/power-pages:force-link-environment` skill wraps the Dataverse `ManageEnvironmentStamp` action (HAR-verified against AppDeploymentConfiguration UI on `supplierportalpipelineshostch.crm17`, 2026-05-11). Implements the [documented Force Link flow](https://learn.microsoft.com/en-us/power-platform/alm/custom-host-pipelines#using-force-link-to-associate-an-environment-with-a-new-host) so users can take over an environment's Pipelines-host association when a previous host has claimed it.
- `setup-pipeline` Phase 5 now detects the *"already associated with another pipelines host"* error and offers the new skill as the documented auto-fix for **Pattern 15** in `references/deployment-error-catalog.md`. Never auto-invoked — gated by explicit `AskUserQuestion` consent because the action is destructive to the previous host (makers there lose pipeline access for the env).
- Drops the `plan-alm` Q7 source-control question + `GIT_STATUS` planData field. It was dead weight (renderer didn't read it, no automation was driven from it). To be re-introduced when the git-versioning skill ships.

## What's in the box

| Layer | Files |
|---|---|
| Skill | `skills/force-link-environment/SKILL.md` + `scripts/validate-force-link.js` (6-phase workflow: prereqs → MCP Learn grounding → identify host+env → resolve record on new host with explicit recovery path when create-helper throws → mandatory destructive-action consent → Force Link + post-validation poll → `.last-force-link.json` marker) |
| Shared lib | `scripts/lib/force-link-environment.js` — CLI + module; HAR-shape body `{ "DeploymentEnvironmentId": "{UPPER-GUID}" }`, required headers `clienthost: Browser`, `x-ms-app-name: AppDeploymentConfiguration`, `prefer: odata.include-annotations="*"`; 204 success; re-polls `validationstatus` to terminal |
| Tests | `scripts/tests/force-link-environment.test.js` (11 unit tests) + `scripts/tests/validate-force-link.test.js` (11 validator tests) + assertion added to `powerpages-hook-utils.test.js` |
| Registration | `scripts/lib/powerpages-hook-utils.js` `TRACKED_SKILLS`; `references/skill-tracking-reference.md` mapping row |
| Docs | `references/deployment-error-catalog.md` → Pattern 15; `references/cicd-pipeline-patterns.md` → new HAR-confirmed `ManageEnvironmentStamp` section; `AGENTS.md`/`CLAUDE.md` skills list; `plugin.json` bumped 1.5.0 → 1.6.0, keywords + description updated |
| setup-pipeline wiring | Phase 5a detects the stderr substring and surfaces the auto-fix offer (no auto-invoke; on user-accepted force-link success, re-runs Phase 5 from the top since the create helper is idempotent) |

## Cleanup riding along

- `plan-alm/SKILL.md` — Q7 git question removed; section headers renumbered (`Q3 through Q7` → `Q3 through Q6` PP; `Q3 through Q8` → `Q3 through Q6` Manual); approval-gate summary lines updated; `GIT_STATUS` removed from `planData` shape and the risks-list rule.
- `scripts/tests/render-alm-plan.test.js` — dead `GIT_STATUS: 'yes'` fixture field removed.
- Verified clean: `git grep "GIT_STATUS\|gitStatus\|git_status\|Git source control\|enabling source control"` returns zero hits.

## Pre-existing nit not fixed here

Manual path in `plan-alm` has a duplicate `**Q6:**` label (the auto-detect env-var step mirrors PP path's Q6 label). Independent of this PR — flagged for whoever re-introduces the git-versioning flow; they'll naturally renumber as they re-insert a question.

## Test plan

- [x] `node --test plugins/power-pages/scripts/tests/` — **690/690 pass** (+24 new tests vs main).
- [x] Lint suite passes (`ManageEnvironmentStamp` is an action invocation, not an entity-set POST, so the manifest rule doesn't fire on the new SKILL.md or catalog entry).
- [x] Force-link helper validates GUID **before** any HTTP call (test asserts `calls === 0` on malformed input).
- [x] Validator blocks on `validationStatus === Failed (200000002)` and on non-terminal Pending — a botched run surfaces in the post-tool hook rather than silently passing.
- [x] `.alm-deferred` marker silent-approves the validator (tested).
- [ ] **Live run pending** — exercise against a tenant where a BAP env is stamped to a different host (Platform Host → Custom Host migration scenario). Not blocking for merge into the ALM feature branch since the API shape is HAR-verified, but worth running before this rolls into `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tyaginidhi:
## Review fixes applied — commit `0fbcc9c`

Pushed in a single follow-up commit. Net diff: **+44 / −16** across 4 files. 690/690 tests still pass.

### Medium

- **#1 Phase 5/5a handoff** — `setup-pipeline` Phase 5 prose now invokes `create-deployment-environment.js` with the actual required args (`--bapEnvId` + `--environmentType`), documents where `bapEnvId` comes from (`pac env list` / `pac env who`), and explicitly tells the agent to retain it for the Phase 5a force-link handoff.
- **#2 Re-run scope** — Phase 5a guardrail tightened: "re-attempt just the failing environment by re-running `create-deployment-environment.js` with the same args" instead of restarting Phase 5 wholesale. The idempotency rationale (`findExistingByBapId` short-circuit returns `reused: true` for already-succeeded envs) is inline.

### Minor

- **#4 API version** — Inline comment in `force-link-environment.js` explains why the post-validation poll uses v9.0 (HAR-captured against `AppDeploymentConfiguration`) while `create-deployment-environment.js` uses v9.1.
- **#5 Substring brittleness** — Phase 5a now lists 6 case-insensitive variants and adds the stable Dataverse error code `0x80048d18` as a fallback signal so backend wording drift doesn't silently break detection.
- **#6 Dead variable** — `lastErrorMessage` scoped to the `VALIDATION_STATUS_FAILED` branch only.
- **#7 ESLint comments** — Replaced `destructure + eslint-disable-line` pattern with a small `withoutField()` helper since the repo doesn't run ESLint.
- **#8 `previousHostEnvUrl` extraction** — SKILL.md Phase 6 now spells out the derivation order: `.last-host-check.json` → errormessage regex → `null` fallback. Marked informational-only (validator already permits `null`).

### Deferred (not in this commit)

- **#3 Initial-sleep penalty** — Left as-is. Matches the existing `create-deployment-environment.js` pattern; deviating would introduce inconsistency for a ~3s/successful-run optimization. Worth revisiting as a cross-cutting cleanup if both helpers ever switch to immediate-first-poll together.

Ready for re-review.

tyaginidhi:
## Env Variables tab UX overhaul — commit \`610787c\`

Pushed as a third commit on this branch. Net diff: **+321 / −30** across 5 files. **693/693 tests pass** (+3 new).

### What's in the commit

- **`discover-env-var-definitions.js`** — surfaces \`displayName\` and \`description\` (previously dropped on the return path even though \`displayname\` was already in the OData \`\$select\`; \`description\` newly added to the projection).
- **\`render-alm-plan.js\`** — replaced the prior flat \`<table>\` with mock-style expandable cards (one per env var); added a per-env values comparison matrix that renders only when at least one var has \`values{}\` populated; tightened the planned-count note from "up to N" to "N candidate site settings one at a time" so users don't read the candidate count as the realized count; appended a neutral refresh hint explaining the tab back-fills after \`setup-solution\` / \`deploy-pipeline\`.
- **\`alm-plan-template.html\`** — new \`.envvar-card\` / \`.envvar-header\` / \`.envvar-body\` / \`.envvar-tag\` / \`.envvar-name\` / \`.envvar-type\` / \`.envvar-chevron\` / \`.envvar-fields\` / \`.envvar-values\` classes modelled on the existing \`.split-solution-card\` pattern; click-toggle JS hook wired alongside the existing solution-card handler.
- **Tests** — three new renderer tests (cards render with the right structure; matrix renders when populated; matrix omitted when no \`values{}\` maps); discovery test extended to assert displayName/description threading + description in the \$select clause.

### Why the live supplier-invoice-portal plan showed an empty tab

Root cause was orchestration, not query — \`size-estimator\` counted 7 env vars with the same prefix filter the discovery uses, but \`plan-alm\` Phase 1 Step 10b either wasn't invoked or its sidecar wasn't ingested into \`planData.envVars\`. I manually re-ran discovery against the dev env + ingested into the project's \`.alm-plan-data.json\`; the regenerated \`alm-plan.html\` now shows 7 expandable cards instead of the "planned (25)" placeholder. The renderer changes here defend the rendering path; a follow-up should make \`setup-solution\` / \`refresh-alm-plan-data\` reliably write \`.last-env-vars.json\` so this populates without manual intervention.

### Scope creep guard

These changes are renderer + discovery only — no new SKILL.md phases, no plan-alm question-flow changes, no setup-solution touches. The Step 10b reliability fix is intentionally **not** in this commit so each fix has a clean review surface.

PR #190, opened 2026-06-15
1DS telemetry: region-aware routing via a per-plugin resolver contract

amitjoshi438:
Adds **region-aware routing** to the (already-shipped) 1DS telemetry — the org's geo (Artemis) + cloud stamp select a per-region iKey/collector — and does it **without putting any routing logic in the shared library**. Region selection lives entirely in the power-pages plugin behind a small generic resolver contract that any plugin can implement (or skip in favor of a single static key).

## What this branch changes 

**Region routing (power-pages-owned)**
- New `plugins/power-pages/scripts/lib/telemetry/region/`: `region-resolver` / `region-cache` / `artemis-service` — org geo via Artemis + cloud stamp → routing region, cached per-org.
- `ikey.json` gains a `default_region` + `regions` map; real per-region iKeys for us/eu/gov/high/dod/mooncake/internal.
- `pac-auth.js` parses the **cloud** stamp from `pac auth who`; `emit-spawn` forwards it to the dispatcher.

**Generic resolver contract (keeps `shared/` routing-agnostic)**
- New `shared/telemetry/lib/resolver-loader.js`: discovers an optional plugin `resolver.js` by convention next to `ikey.json`.
- `emit-dispatcher.js` resolves the destination by precedence — env override (test seam) → plugin `resolver.js` → static `instrumentationKey`/`collector_url` → none — and contains **no** region/artemis code.
- Generic provisioning fast-gate via `resolver.isProvisioned(cfg)` (static-key fallback) in `emit-from-prompt` + the pretool hook; both forward `ikeyJsonPath` so the detached dispatcher reads the plugin config.
- `plugins/power-pages/scripts/lib/telemetry/resolver.js` adapts the region code to the contract (`resolve()` / `isProvisioned()`).


plugins/power-pages/scripts/lib/telemetry/region/region-cache.js
```diff
@@ -0,0 +1,72 @@
...
+    iKey: entry.iKey,
+    collectorUrl: entry.collectorUrl,
+  };
+}
+
+function write(orgId, entry, configDir) {
+  if (!orgId || !entry) return;
+  const dir = configDir || defaultDir();
+  try {
+    fs.mkdirSync(dir, { recursive: true });
```
priyanshu92, line 58: write() reads the whole region-cache.json, mutates it in memory, then fs.writeFileSync()s the entire file. Each tracked-skill invocation spawns a separate detached dispatcher (emit-spawn.js fireAndForget), so back-to-back skill calls (or parallel sessions / multiple plugins sharing ~/.power-platform-skills/region-cache.json) can race: A reads {orgA}, B reads {orgA}, A writes {orgA, orgB}, B writes {orgA} — A's orgB entry is lost. The read path tolerates malformed JSON (returns null), so a torn write only causes a cache miss; the real cost is losing already-resolved entries, forcing another Artemis round-trip (up to the full 5s timeout on slow networks) and undermining the 24h TTL.

amitjoshi438: Proposed redesign to close both:

1. Cache the org→region mapping only ({ region, expiresAt }); map region→iKey from the calling plugin's own regionsMap on each hit (free, already in memory). The Artemis round-trip is the only expensive part, and region is plugin-independent — so the cache becomes correctly shareable across plugins instead of a misrouting hazard.
2. Per-org files (region-cache/<orgId>.json, atomic write each) — removes the shared read-modify-write entirely, so the cross-org lost-update can't happen.
3. Only cache when region was derived from a recognized geo/stamp, not a per-plugin defaultRegion fallback.

amitjoshi438: Implemented

priyanshu92 (approved):

PR #195, opened 2026-06-17
Per-plugin telemetry env-var toggle + enable power-pages transmission

amitjoshi438:
- Adds a **per-plugin telemetry env-var toggle** `POWER_PLATFORM_SKILLS_TELEMETRY_<PLUGIN>` (power-pages: `POWER_PLATFORM_SKILLS_TELEMETRY_POWER_PAGES`, values `on|off`) so automation/CI can toggle 1DS telemetry without the slash skill. Mirrors `/telemetry off`: suppresses **transmission only** — the local `events.jsonl` mirror is still written.
- **Precedence:** a persisted `config.json` per-plugin choice always wins; the env var applies only when no choice is stored; otherwise default-on.
- All logic in shared `user-config.js` (`telemetryEnvVarName`, `readTelemetryEnvChoice`, `effectiveTelemetryChoice`), consumed by both the dispatcher and the CLI (DRY). Dispatcher needs no source change — reads the override via `isTransmissionOptedOut`'s default `process.env`, after the mirror write.
- **Enables power-pages transmission**: flips `ikey.json` `disabled: false`. Verify the tenant-side Kusto stream + annotation for `PagesAIPluginEvent` are provisioned.
- Docs updated across plugin/root README, shared telemetry README, the telemetry workflow doc, and both AGENTS.md telemetry sections.


plugins/power-pages/README.md
```diff
@@ -399,6 +399,7 @@ Enables, disables, or checks the status of anonymous usage telemetry. Per-user a
 - `/power-pages:telemetry off` — stop sending telemetry (nothing leaves your machine)
 - `/power-pages:telemetry on` — resume sending telemetry
 - No personal data is ever collected (anonymous: skill name, plugin version, OS, Node version)
+- Automation/CI: set `POWER_PLATFORM_SKILLS_TELEMETRY_POWER_PAGES=off` (a saved choice wins over the env var)
```
priyanshu92: Let's change the env var to `POWER_PLATFORM_SKILLS_TELEMETRY_POWER_PAGES_OPTOUT`. If the variable is set then telemetry should be disabled. This follow the patterns of some other CLIs like `dotnet`. Also, if telemetry is on using the skills but this env var is set then it should be disabled. Env vars have highest precedence everywhere.

amitjoshi438: Okay, updated the design and implementation.

priyanshu92 (approved):

PR #207, opened 2026-06-22
Fix plugin root resolution

priyanshu92:
- Resolve plugin root paths in hooks and MCP launchers without relying on placeholder expansion.
- Keep marketplace plugin entries minimal while preserving marketplace-level metadata.
- Bundle shared workflow and telemetry runtime files as physical plugin content instead of symlinks.


tyaginidhi (commented):
# Code Review — Fix plugin root resolution

**+2567 / −285 across 37 files · CI: ✅ all green**

## Overview
Fixes unreliable plugin-root resolution by removing two mechanisms that don't survive Windows / plugin-host installs:
1. **`${CLAUDE_PLUGIN_ROOT}` placeholder** in `.mcp.json` + `hooks.json` → replaced with an inline `node -e` that resolves `PLUGIN_ROOT || CLAUDE_PLUGIN_ROOT` from the environment and `require()`s the target.
2. **Git symlinks** for shared content (telemetry runtime, shared workflow `.md`s) → replaced with committed **physical copies**.
3. Bonus: marketplace `plugins[]` entries trimmed to **minimal** (`name`+`source`), hardened telemetry `user-config` against a malformed `telemetry: []`, and corrected the privacy disclosure.

## Correctness — verified sound ✅
- **MCP launcher** works: `launch-playwright-mcp.js` already exports `launch` behind a `require.main === module` guard, so `node -e "...mod.launch()"` correctly runs it (the guarded CLI block doesn't double-fire).
- **Hook launchers** work: all three hook scripts execute at top-level (IIFE / `process.stdin` handlers), so `require()` runs them. Hooks **fail-open** (`exit 0` when root unset) — correct posture for telemetry/validation; MCP **hard-fails** (`throw`) — also correct.
- **Physical copies are byte-identical** to the *edited* `shared/telemetry/lib` source — copies were correctly refreshed after the shared edits (checked `emit-spawn`, `telemetry-config`, `user-config`, `events`, `agent-info`).
- **Privacy disclosure is now accurate**: `events.js` already allowlists `orgId`/`tenantId`/`aiAgentVersion`, so the *old* text ("never includes … hostnames") understated collection. The new wording fixes a real accuracy gap. 👍
- Convention docs (`AGENTS.md`, and `CLAUDE.md` → symlink to it) are updated consistently, including the *why* ("Windows and plugin-host installs can materialize symlinks as plain link files").

## Issues & risks

- **🔶 Main concern — no drift protection for the physical copies.** The symlink *couldn't* drift; a committed copy can. `AGENTS.md` now says "refresh every adopting plugin's copy in the same change," but **nothing enforces it** — there's no test/CI asserting `plugins/*/scripts/lib/telemetry/lib` equals `shared/telemetry/lib` (same for bundled workflow `.md`s). The next person who edits `shared/` and forgets to re-copy ships stale plugin code with green CI.
  - **Suggestion:** add a cheap guard — `readdirSync` over `shared/telemetry/lib` and assert each file is byte-identical to the adopting plugin's copy (and the workflow `.md`s). This directly catches the failure mode the new manual process invites. Highest-value addition.

- **🔹 Duplicated inline `node -e` blob.** The resolver one-liner is copy-pasted across 2 `.mcp.json` + 3 `hooks.json` entries with no test. If the resolution logic ever changes, it's edited in many places. Acceptable as a workaround, but a small maintenance smell.

- **🔹 Coordination with #202.** This PR **removes `version` from `marketplace.json` plugin entries** entirely — which makes the marketplace↔plugin version-mirror sync that #202/#194 maintain moot. Whichever merges second needs a rebase: #202 still bumps a `version` field this PR deletes. Not a defect in #207, but flag the ordering so the legacy-compat validator doesn't fail on the loser.

- **🔹 Strictness of `assertMinimalMarketplaceEntry`.** It enforces *exactly* `['name','source']` — intentional per the Open Plugins rationale, but it will hard-reject any future legitimate marketplace-level override field. Fine as a deliberate gate; just noting it's strict.

## Conventions / tests / security
- **Conventions:** ✅ Docs and validator updated to match; no doc/code contradiction left behind.
- **Tests:** ✅ The `user-config` hardening has a dedicated new test (`telemetry: []` normalization). ❌ Gap: no test for copy↔shared parity (see main concern).
- **Security/privacy:** ✅ Net-positive — disclosure now matches the allowlist; org/tenant collection is pre-existing (not introduced here) and governed by the events allowlist/scrubber. `node -e` runs only trusted committed paths.

## Verdict
**Solid, root-cause fix — approve once the drift concern is addressed.** The launcher and copy mechanics are correct and verified, the privacy-disclosure fix is a genuine improvement, and the docs are kept honest. The one thing I'd ask for before merge is a **copy-parity test** so the de-symlinked content can't silently drift; the #202 version-field coordination is a heads-up rather than a blocker.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

priyanshu92:
> # Code Review — Fix plugin root resolution
> **+2567 / −285 across 37 files · CI: ✅ all green**
> 
> ## Overview
> Fixes unreliable plugin-root resolution by removing two mechanisms that don't survive Windows / plugin-host installs:
> 
> 1. **`${CLAUDE_PLUGIN_ROOT}` placeholder** in `.mcp.json` + `hooks.json` → replaced with an inline `node -e` that resolves `PLUGIN_ROOT || CLAUDE_PLUGIN_ROOT` from the environment and `require()`s the target.
> 2. **Git symlinks** for shared content (telemetry runtime, shared workflow `.md`s) → replaced with committed **physical copies**.
> 3. Bonus: marketplace `plugins[]` entries trimmed to **minimal** (`name`+`source`), hardened telemetry `user-config` against a malformed `telemetry: []`, and corrected the privacy disclosure.
> 
> ## Correctness — verified sound ✅
> * **MCP launcher** works: `launch-playwright-mcp.js` already exports `launch` behind a `require.main === module` guard, so `node -e "...mod.launch()"` correctly runs it (the guarded CLI block doesn't double-fire).
> * **Hook launchers** work: all three hook scripts execute at top-level (IIFE / `process.stdin` handlers), so `require()` runs them. Hooks **fail-open** (`exit 0` when root unset) — correct posture for telemetry/validation; MCP **hard-fails** (`throw`) — also correct.
> * **Physical copies are byte-identical** to the _edited_ `shared/telemetry/lib` source — copies were correctly refreshed after the shared edits (checked `emit-spawn`, `telemetry-config`, `user-config`, `events`, `agent-info`).
> * **Privacy disclosure is now accurate**: `events.js` already allowlists `orgId`/`tenantId`/`aiAgentVersion`, so the _old_ text ("never includes … hostnames") understated collection. The new wording fixes a real accuracy gap. 👍
> * Convention docs (`AGENTS.md`, and `CLAUDE.md` → symlink to it) are updated consistently, including the _why_ ("Windows and plugin-host installs can materialize symlinks as plain link files").
> 
> ## Issues & risks
> * **🔶 Main concern — no drift protection for the physical copies.** The symlink _couldn't_ drift; a committed copy can. `AGENTS.md` now says "refresh every adopting plugin's copy in the same change," but **nothing enforces it** — there's no test/CI asserting `plugins/*/scripts/lib/telemetry/lib` equals `shared/telemetry/lib` (same for bundled workflow `.md`s). The next person who edits `shared/` and forgets to re-copy ships stale plugin code with green CI.
>   
>   * **Suggestion:** add a cheap guard — `readdirSync` over `shared/telemetry/lib` and assert each file is byte-identical to the adopting plugin's copy (and the workflow `.md`s). This directly catches the failure mode the new manual process invites. Highest-value addition.
> * **🔹 Duplicated inline `node -e` blob.** The resolver one-liner is copy-pasted across 2 `.mcp.json` + 3 `hooks.json` entries with no test. If the resolution logic ever changes, it's edited in many places. Acceptable as a workaround, but a small maintenance smell.
> * **🔹 Coordination with #202.** This PR **removes `version` from `marketplace.json` plugin entries** entirely — which makes the marketplace↔plugin version-mirror sync that #202/#194 maintain moot. Whichever merges second needs a rebase: #202 still bumps a `version` field this PR deletes. Not a defect in #207, but flag the ordering so the legacy-compat validator doesn't fail on the loser.
> * **🔹 Strictness of `assertMinimalMarketplaceEntry`.** It enforces _exactly_ `['name','source']` — intentional per the Open Plugins rationale, but it will hard-reject any future legitimate marketplace-level override field. Fine as a deliberate gate; just noting it's strict.
> 
> ## Conventions / tests / security
> * **Conventions:** ✅ Docs and validator updated to match; no doc/code contradiction left behind.
> * **Tests:** ✅ The `user-config` hardening has a dedicated new test (`telemetry: []` normalization). ❌ Gap: no test for copy↔shared parity (see main concern).
> * **Security/privacy:** ✅ Net-positive — disclosure now matches the allowlist; org/tenant collection is pre-existing (not introduced here) and governed by the events allowlist/scrubber. `node -e` runs only trusted committed paths.
> 
> ## Verdict
> **Solid, root-cause fix — approve once the drift concern is addressed.** The launcher and copy mechanics are correct and verified, the privacy-disclosure fix is a genuine improvement, and the docs are kept honest. The one thing I'd ask for before merge is a **copy-parity test** so the de-symlinked content can't silently drift; the #202 version-field coordination is a heads-up rather than a blocker.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Will do this in a later PR

tyaginidhi (approved):

PR #225, opened 2026-07-06
Enforce kebab-case for marketplace keywords/tags

priyanshu92:
## What & why

Ensures all `keywords` (and future `tags`) across the marketplace files are in **kebab-case**, and adds CI so PRs fail when they are not.

### Changes
- **Fix `mobile-apps` keywords** — the only manifest with non-kebab-case keywords (spaces). Updated both mirrors (`.plugin/plugin.json` and `.claude-plugin/plugin.json`):
  - `power platform` → `power-platform`
  - `power apps` → `power-apps`
  - `code apps` → `code-apps`
  - `react native` → `react-native`
- **Add `scripts/validate-keyword-case.js`** — scans `marketplace.json`, its legacy `.claude-plugin/marketplace.json` mirror, and every plugin manifest (both `.plugin` and `.claude-plugin`). Validates `keywords` and `tags` (top-level and per-plugin entries) against `^[a-z0-9]+(-[a-z0-9]+)*$`, rejecting spaces, uppercase, underscores, dots, and leading/trailing/double hyphens. Exits non-zero with per-entry error messages. `tags` is covered proactively (none exist today).
- **Add `.github/workflows/validate-keyword-case.yml`** — runs the validator on every `pull_request` to `main` (mirrors the existing `validate-repository-metadata.yml` style), so a non-kebab-case keyword/tag fails the PR.

### Verification
- Validator passes across all 14 marketplace files after the fix.
- Confirmed it fails on injected bad terms (`Bad Keyword`, `under_score`, `trailing-`, and a non-kebab `tags` entry).
- `node scripts/validate-legacy-compatibility.js` still reports the mirrors in sync.

### DIFF
--- a/.github/workflows/validate-keyword-case.yml
+++ b/.github/workflows/validate-keyword-case.yml
@@ -1,0 +1,23 @@
+name: validate-keyword-case
+
+on:
+    pull_request:
+        branches:
+            - main
+    workflow_dispatch:
+
+jobs:
+    validate-keyword-case:
+        name: validate-keyword-case
+        runs-on: ubuntu-latest
+        steps:
+            - name: checkout
+              uses: actions/checkout@v4
+
+            - name: setup-node
+              uses: actions/setup-node@v4
+              with:
+                  node-version: 20
+
+            - name: validate-keyword-tags-kebab-case
+              run: node scripts/validate-keyword-case.js
--- a/plugins/mobile-apps/.claude-plugin/plugin.json
+++ b/plugins/mobile-apps/.claude-plugin/plugin.json
@@ -10,12 +10,12 @@
     "repository": "https://github.com/microsoft/power-platform-skills/",
     "license": "MIT",
     "keywords": [
-        "power platform",
-        "power apps",
-        "code apps",
+        "power-platform",
+        "power-apps",
+        "code-apps",
         "mobile",
         "expo",
-        "react native",
+        "react-native",
         "native",
         "connectors",
         "microsoft"
--- a/plugins/mobile-apps/.plugin/plugin.json
+++ b/plugins/mobile-apps/.plugin/plugin.json
@@ -10,12 +10,12 @@
     "repository": "https://github.com/microsoft/power-platform-skills/",
     "license": "MIT",
     "keywords": [
-        "power platform",
-        "power apps",
-        "code apps",
+        "power-platform",
+        "power-apps",
+        "code-apps",
         "mobile",
         "expo",
-        "react native",
+        "react-native",
         "native",
         "connectors",
         "microsoft"
--- a/scripts/validate-keyword-case.js
+++ b/scripts/validate-keyword-case.js
@@ -1,0 +1,117 @@
+#!/usr/bin/env node
+
+/**
+ * Validates that every `keywords` and `tags` entry across the marketplace files
+ * is in kebab-case. Keywords/tags are the discovery surface for the marketplace,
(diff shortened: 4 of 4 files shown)


tyaginidhi:
## 🤖 dev-team reviewer — PR #225 review

**Verdict: ✅ Safe to merge.** 5 findings — 0 blocking · 1 medium · 2 low · 2 nit. Verified against a live checkout: validator passes on all 14 marketplace files, legacy mirrors in sync (`validate-legacy-compatibility.js` passes), and malformed JSON exits non-zero (fails safe).

---

### 🟡 Medium — Malformed JSON produces a raw stack trace instead of a clean per-file error
`scripts/validate-keyword-case.js` (`readJson`)

```js
function readJson(filePath) {
  return JSON.parse(fs.readFileSync(filePath, 'utf8'));  // throws uncaught
}
```

Pointing at an invalid manifest yields a Node `SyntaxError` dump with **no indication of which file failed**. It still fails CI (fails safe — hence medium, not high), but under a workflow named `validate-keyword-case` a plain JSON typo produces a misleading, file-less error. Suggested fix: wrap the read and report through the same per-entry channel, e.g.
```js
try { document = readJson(filePath); }
catch (e) { errors.push(`${relative(filePath)}: invalid JSON — ${e.message}`); return; }
```

### ⚪ Low — Keywords/tags nested under `metadata` would silently escape validation
Only top-level `document.keywords`/`tags` and `plugins[].keywords`/`tags` are checked. `marketplace.json` stores a `metadata` object; if discovery terms are ever added under `metadata.keywords`, they'd be missed. None exist today — purely forward-looking hardening to make the "future additions can't slip through" comment fully true.

### ⚪ Low — No `paths:` filter and no explicit `permissions:` block
`.github/workflows/validate-keyword-case.yml` runs on every PR to `main` with default token permissions. This is **consistent with the existing `validate-repository-metadata.yml`**, so it matches house style — only worth tightening (`permissions: contents: read`) if standardizing least-privilege across all workflows at once. Not a PR-specific defect.

### ⚪ Nit — Telemetry opt-out correctly omitted
This job runs only the pure-JS validator (no telemetry-emitting Power Pages hooks), so omitting `POWER_PLATFORM_SKILLS_TELEMETRY_POWER_PAGES_OPTOUT` is correct and matches `validate-repository-metadata.yml`. No change needed.

### ⚪ Nit — mobile-apps keyword changes improve consistency (not drift)
Cross-checked siblings: `power-platform` matches `code-apps`; `power-apps` matches `canvas-apps`/`code-apps`; `code-apps` matches `code-apps`; `react-native` is new but well-formed with no conflict. The renames **reduce** near-duplicate variance. (Pre-existing, unrelated: `mobile-apps` is the only plugin with the bare `microsoft` keyword — untouched here; possible follow-up cleanup.)

---

### Verified correct
- Regex `^[a-z0-9]+(?:-[a-z0-9]+)*$` — rejects spaces, uppercase, underscores, dots, empty, and leading/trailing/double hyphens; accepts `spa`, `web-api`, `react-native`.
- Coverage — exactly the 12 keyword-bearing plugin manifests + 2 marketplace files = 14, no gaps.
- Both mobile-apps mirrors updated identically and in sync.
- Non-array/non-string guards and the `null`/non-object plugin guard are sound; deterministic ordering via `.sort()`.

Recommendation: merge as-is; optionally address the 🟡 medium (JSON error message) here or as a fast follow-up for better contributor DX.

tyaginidhi (approved):

PR #275, opened 2026-07-20
Require structured GenPage agent messages

nguyentony95:
## Summary
- require GenPage create, edit, and navigation-fix uploads to send structured `Agent Thoughts`, `Summary`, and `Final Code` sections
- use literal `\n` sequences so PAC on Windows receives `--agent-message` as a single argument
- require concise user-facing rationale rather than private chain-of-thought or sensitive information
- enforce the contract for every initial upload and re-upload in Layer 1 evals

## Runtime dependency
Rendering escaped structured sections requires the GenUX escaped-section support deployed with `MscrmControls.UxAgentControl` version `1.0.285`. Merge/release this instruction change only when that runtime support is available in the target environments.

## Validation
- model-apps script tests
- GenPage eval unit tests
- Layer 1 smoke fixtures
- Layer 2 smoke fixtures


akshaymaloo (changes requested):
I don't think we need this anymore?

PR #291, opened 2026-07-23
code-apps: add Work IQ (M365 Copilot Search) skill

madhumach:
## What

Adds a new `/add-workiq` skill to the **code-apps** plugin for integrating **Work IQ (M365 Copilot Search)** via the Work IQ Copilot MCP connector (`shared_a365copilotchatmcp`).

### Changes
- **NEW** `skills/add-workiq/SKILL.md` — connector setup (`create-connection` + `add-data-source`) plus a production-ready `McpSession` wrapper for the stateful MCP protocol.
- `skills/add-connector/SKILL.md` — dedicated-skill entry + delegation row for `shared_a365copilotchatmcp`.
- `agents/code-app-architect.md` — connector table row + workaround note.
- `skills/create-code-app/SKILL.md` — Work IQ added to connector selection lists.

### Key technical detail
The connector's MCP server (`Microsoft.MCPPlatform.WebApi`) is **stateless-tolerant**: the client must **not** send an `Mcp-Session-Id` on `initialize` (doing so returns `-32001 Session not found`). The `McpSession` wrapper handles this, sequences JSON-RPC ids, auto-retries, and parses the nested response.

### Validation
Validated **end-to-end** against a freshly scaffolded code app: `create-connection` → `add-data-source` → `McpSession` wrapper → `CopilotChat` returned a real M365-grounded answer with citations. Repo validators (skill-descriptions, keyword-case, plugin-names) pass locally.

Screenshots:
1)
image


pavankm:
Can we delete the screenshots? They contain email IDs and internal links

plugins/code-apps/skills/add-workiq/SKILL.md
```diff
@@ -0,0 +1,480 @@
...
+```typescript
+import { McpSession } from './connectors/mcpClient'
+
+// Initialize once per app session and reuse across calls.
+const workIqSession = new McpSession()
+
+export async function queryWorkIQ(userPrompt: string): Promise<string> {
+  try {
+    const { text } = await workIqSession.callCopilotChat(userPrompt)
+    if (text.startsWith('Error:')) {
```
pavankm: This is a bit fragile, could we throw an error in callCopilotChat instead and catch it?

plugins/code-apps/skills/add-workiq/SKILL.md
```diff
@@ -0,0 +1,480 @@
...
+    const negotiatedSessionId = this.extractSessionId(raw)
+    if (negotiatedSessionId) {
+      this.sessionId = negotiatedSessionId
+    }
+
+    const parsed = parseRpc(raw)
+
+    if (allowRetry && method !== 'initialize' && this.isSessionNotFound(parsed)) {
+      this.resetSession()
+      await this.initialize()
```
pavankm, line 245: Since we aren't throwing an error, if initialization fails this just continues to the send method.
It would be better to fail fast and surface the initialization error

pavankm (approved):

PR #309, opened 2026-07-26
Add manage-governance Power Pages skill

kumarsach07:
Adds a new **Power Pages** plugin skill, `manage-governance`, to apply and inspect env-level Power Pages governance policies (Maker Copilot + authentication-related Enable* toggles). The PR introduces the skill workflow/spec, supporting scripts for listing envs/portals, applying policies with polling + verification, deterministic consent-gate rendering, and a supporting reference mapping + test suite.

**Changes:**
- Introduces the `manage-governance` skill spec (`SKILL.md`), reference data (`governance-mapping.json`, training cases), and end-to-end documentation.
- Adds gateway-based scripts to list environments/sites, parse portal input, render tables/impact summaries, and apply governance policies with polling.
- Adds node:test coverage for key renderers/mappings/payload logic and updates Power Pages plugin docs to include the new skill.

### Policy status env
image

Impact Summary 
image
image


kumarsach07:
Pushed a fix addressing review defects **1 (dead-end OR remediation)** and **2 (intent not resumed)** — both cured by one mechanism.

**Resolver (
esolve-portal-availability.js)**
- Added 	ransitiveParents, vailabilityDepth, computeRemediationChain.
- computeAvailability now emits an ordered **
emediationChain** (currently-disabled gating parents, **root-first** via the DAG) and **
ext** (= chain[0]).

**Defect 1** — the orchestrator now offers **only 
ext**, never a flat "External Auth **or** OAuth 2.0". Because the chain is root-first, 
ext is always the parent that can actually be enabled now (External Auth before OAuth 2.0), so the second dead-end is gone.

**Defect 2** — SKILL.md persists `PENDING_CHILD_INTENT` and auto-re-presents the original child (e.g. "enable Google") after each parent is enabled (new **Step B.2**), wired into the named-site path (4.2.2b) and the Phase 5 loop. Parents keep their own sign-out consent gate — never silently auto-enabled.

**Tests/validation** — new `resolve-portal-availability-chain.test.js` (9 cases: root-first ordering, single/two-parent, one-parent-already-on advance, unread-parent fail-open, leaf). Full suite **1332 green**, `alm-lint` 0 findings.

The non-blocking architecture follow-up (extract auth/ring + portal listing + `table.js` into `scripts/lib/`) is tracked as a separate fast-follow.

kumarsach07:
Addressed the cheap batch of automated review comments in `e4a6dc3d`:

- **buildRoute `/governance/undefined`** — `buildRoute()` now fails fast when a policy-encoding op is missing `policy` (and `getPortal` missing `portalId`), instead of silently 404-ing on a bad URL. `apply` stays exempt (policy is in the POST body).
- **`mdCell()` newline handling** — now flattens CR/LF runs to a space so a newline in an env display name can't break the Markdown table row.
- **`list-portals.js` comment** — corrected: the client sends plain `select`/`skip` query keys, not OData `$select`/`$skip`.
- **`AGENTS.md` manage-governance description** — fixed: portal-scoped apply POSTs `/governance` (differing only by body), and polling is every 30s, not 5s.

Added `governance-route-guards.test.js` (10 cases) covering the new throw paths and the newline flattening. Full CI suite **1342 pass / 0 fail**; `alm-lint` 0 findings.

The remaining review comments (`PP_GOV_API_HOST` token, `tip-auth` tmp-cache location, and the `scripts/test/governance/` payload/tip-auth tests) are tracked separately — the latter live in a directory the CI workflow does not execute, so they are not build-breaking.

PR #313, opened 2026-07-27
perf(mobile-apps): speed up and harden Dataverse schema creation

kanu-shubham:
## Summary

This PR improves Dataverse schema-provisioning performance while preserving the same generated data model, relationships, security behavior, and offline capabilities.

The optimized workflow created 8 tables and 10 lookup relationships in **4 minutes 17 seconds**, compared with **5 minutes 56 seconds** for the baseline.

## What changed

The Dataverse creation workflow was streamlined to reduce metadata-provisioning overhead:

- Creates each table with its planned non-lookup columns included in the table request.
- Creates lookup relationships only after both endpoint tables exist.
- Preserves dependency-tier ordering.
- Avoids unsafe concurrent metadata writes, which can cause Dataverse metadata-lock failures.
- Retains collision detection, idempotency checks, solution targeting, offline availability, and change tracking.
- Does not remove or simplify any tables, columns, or relationships to achieve the improvement.

The optimization affects provisioning time only. The resulting Dataverse schema remains functionally equivalent.

## Benchmark scope

Both runs used the same:

- Dataverse environment
- Default solution and publisher
- 8 new tables
- 10 one-to-many lookup relationships
- Column types, including text, Choice, Yes/No, Date/Time, Decimal, Memo, and Image
- Dependency and creation order
- Offline and change-tracking configuration

Fresh logical names were used for the optimized run to avoid idempotent skips and deleted-table tombstones.

## Results

| Metric | Baseline | Optimized | Improvement |
|---|---:|---:|---:|
| Tables created | 8 | 8 | Same scope |
| Relationships created | 10 | 10 | Same scope |
| Total metadata operations | 18 | 18 | Same scope |
| Total duration | 5m 56.1s | 4m 16.8s | **1m 39.2s faster** |
| Average time per operation | 19.8s | 14.3s | **5.5s faster** |
| Observed performance improvement | — | — | **27.9%** |
| Throughput increase | — | — | **1.39×** |

All 8 optimized-run tables were queried after creation and confirmed to exist in Dataverse.

## Throttling-adjusted comparison

The baseline encountered one 30-second Dataverse throttling delay. The optimized run did not show a comparable delay.

After excluding that delay:

| Metric | Result |
|---|---:|
| Adjusted baseline | 5m 26.1s |
| Optimized run | 4m 16.8s |
| Adjusted saving | **1m 9.2s** |
| Adjusted improvement | **21.2%** |

This indicates that the result is not explained only by the throttling event, although Dataverse service variability can affect individual runs.

## Leadership impact

The observed improvement reduces developer waiting time during:

- Initial application provisioning
- Dataverse data-model iteration
- Automated environment setup
- Benchmark and test-environment creation
- Repeated development and CI workflows

Projected savings at the observed rate:

| Provisioning runs | Approximate time saved |
|---:|---:|
| 10 | 16m 32s |
| 50 | 1h 22m 42s |
| 100 | 2h 45m 25s |

## Safety and compatibility

The performance improvement does not trade away correctness:

- Dataverse metadata writes remain sequential.
- Parent tables are created before dependent tables.
- Relationships are created only after both tables are available.
- All artifacts remain targeted to the expected solution.
- Tables remain enabled for offline use and change tracking.
- Choice, lookup, image, date, and decimal metadata remain unchanged.
- No existing tables or data were deleted or modified.

## Validation

- 8 of 8 tables created successfully.
- 10 of 10 lookup relationships created successfully.
- All created tables returned successful metadata queries.
- No sample records were inserted as part of the benchmark.
- No table-creation failures were observed in the optimized run.

## Conclusion

The initial like-for-like benchmark shows that the optimized Dataverse provisioning workflow is **approximately 28% faster**, or **approximately 21% faster after accounting for baseline throttling**.

This is a meaningful improvement for developer productivity and automated provisioning while retaining equivalent schema output and Dataverse safety controls.

Because this comparison currently represents one run per variant, the result should be treated as an initial performance signal. Repeated runs should use the median and slowest execution times for a production performance commitment

Cut wall-clock time out of the mobile-app Dataverse workflow, and add a fail-closed reconciliation step before any metadata write.

Three independent speedups:

1. **Auth short-circuit** — stop spawning `az account show` on every Dataverse call.
2. **Batched metadata reads** — replace `2N` per-table discovery requests with one filtered query.
3. **Create-path shape** — one POST per new table with all ordinary columns inline; one column snapshot per existing table.

## 1. Auth Short-Circuit

`getAuthToken()` built its tenant candidates as an array literal:

```js
const tenantCandidates = [
  process.env.POWER_PLATFORM_TENANT_ID,
  process.env.DATAVERSE_TENANT_ID,
  await getDataverseTenantFromChallenge(resourceUrl),  // HTTPS probe
  getAzAccountTenantId(),                              // az account show
].filter(Boolean);
```

A JavaScript array literal evaluates **every element eagerly**, before `.filter()` runs. So the challenge probe and `az account show` executed on **every single call**, even when an env var already supplied the tenant and would have won the loop.

Measured on a warm macOS box:

| Operation | Time |
|---|---|
| `az account show` | **4.857s** |
| `az account get-access-token` (warm) | 0.349s |
| Node startup | 0.107s |

Candidates are now produced lazily and the loop stops at the first token. **Preference order, de-duplication, and the unqualified-token fallback are unchanged** — only redundant work is skipped. For reference, `power-pages` issues a single `az account get-access-token` per call and pays ~0.46s of overhead.

`/add-dataverse` now also exports `POWER_PLATFORM_TENANT_ID` once from the value `resolve-environment.js` already returns, so later calls short-circuit at the first candidate and skip the challenge probe entirely.

## 2. Batched Metadata Reads

Discovery previously issued one entity GET plus one attributes GET per planned table. Both collapse into a single documented query:

```http
GET EntityDefinitions
  ?$select=...
  &$filter=LogicalName eq a or LogicalName eq b
  &$expand=Attributes($select=...)
```

This is the [documented way to query multiple table definitions at once](https://learn.microsoft.com/power-apps/developer/data-platform/query-schema-definitions#basic-retrievemetadatachanges-example). A name present in `value[]` exists; a name absent does not — which is exactly the signal the reconciliation matrix needs. The expanded `$select` stays on base `AttributeMetadata` properties because one query [cannot cast to a derived column type](https://learn.microsoft.com/power-apps/developer/data-platform/query-schema-definitions#evaluate-other-options-to-retrieve-schema-definitions).

Two follow-on reductions fall out of that:

- **Step 5a** now reads the Step 4 snapshot instead of re-probing each create target. It only re-queries a name the batch did not cover, such as a rename candidate generated later.
- **Step 6c** verifies all created tables with one filtered query instead of one GET per table.

Tombstones and hidden collisions were never reliably visible to the pre-flight anyway — the POST-time collision rescue remains the real safety net, and is unchanged.

## 3. Create-Path Shape

**New tables:** one `EntityDefinitions` POST carrying the primary name plus every planned ordinary column, guarded by a five-point self-check (completeness, no inline lookup/customer/owner metadata, no deferred or server-owned columns, exactly one primary-name attribute, no duplicates). Lookups and relationships are a deliberate second pass.

**Existing tables:** compare against the cached snapshot, then POST only missing columns sequentially — replacing one preflight GET per planned column.

Metadata writes stay strictly sequential (Dataverse serializes them behind an exclusive lock) and no metadata `$batch` is introduced.

## Reconciliation Before Writes

Every planned table and column is classified against live target metadata as `reuse`, `extend`, `create`, or `block`, using `IsManaged`, `IsCustomizable`, and `CanCreateAttributes`. A global barrier requires zero blocked items before Step 5 begins.

Missing standard or managed dependencies are `block` with remediation to install/import the owning solution — never recreated as a custom imitation. `replace` is deliberately not an automatic state, since Dataverse cannot change a column type in place.

Planning stays non-blocking: if discovery cannot run, the architect still drafts a plan, marks it `Discovery skipped`, flags entities `Unverified`, and returns `DONE_WITH_CONCERNS`. `/add-dataverse` re-queries live metadata and refuses to write what it cannot confirm.

## No New Validator Script

An earlier revision added a `validate-table-create-payload.js` validator. It was removed: the expected-column list came from the same agent that produced the payload, so it could not independently prove plan alignment, and it compared only names — not types, formats, or lengths. The rule is now a self-check in the skill. `dataverse-request.js` remains the only script in that path.

## Estimated Impact

For a model with ~7 planned tables (5 new, 2 extended) and 6 lookups, Dataverse calls drop from roughly **45 to 21**. Combined with removing `az account show` from every call, per-call overhead falls from ~5.6s to ~0.5s.

Derived from the measured per-operation costs above rather than an end-to-end benchmark, that is on the order of **3–4 minutes** of overhead removed per data-model run. The column-inlining change additionally removes up to `C-1` writes per new table when the agent would otherwise have created a table shell.

## Files

- `plugins/mobile-apps/scripts/lib/validation-helpers.js` — lazy tenant resolution.
- `plugins/mobile-apps/scripts/tests/auth-token-resolution.test.js` — new regression tests.
- `plugins/mobile-apps/skills/add-dataverse/SKILL.md` — batched discovery, snapshot-based pre-flight, batched verify, reconciliation matrix, write barrier, create self-check, two-pass ordering.
- `plugins/mobile-apps/agents/data-model-architect.md` — batched target query, target-grounded classification, non-blocking discovery failures.

## Validation

- `node --test plugins/mobile-apps/scripts/tests/` — **19 passed**
- **Regression proven:** reverting `getAuthToken` to the eager version makes the new `no az account show` test **fail**; the fix makes it pass.
- All 8 consumer scripts parse; module exports and the async `getAuthToken` signature unchanged.
- `validate-skill-descriptions` / `ensure-skill-version-check` / `validate-keyword-case` / `validate-plugin-names` / `validate-legacy-compatibility` / `validate-telemetry-ikeys` — all pass.
- `git diff --check` — clean.

No live Dataverse environment was mutated during validation.

## Compatibility and Risk

- **Auth:** behavior-preserving — same candidate order, same de-duplication, same fallback. Multi-tenant users still receive tenant-targeted tokens. `getAuthToken` remains re-callable for 401 refresh (no memoization).
- **Batched reads:** uses documented OData syntax. If a `$filter` URL would grow impractically long, the skill splits it into a few queries — still far fewer than one per table.
- **Mutation:** more conservative — unverified or incompatible schema now blocks instead of proceeding.
- **Planning:** unchanged in availability; a plan is still produced when discovery fails, just marked unverified.
- Existing-table column creation is still one POST per missing column.
- Live-environment smoke validation is still recommended for new, extend, managed-dependency, and collision cases.

## Rollback

Revert the commits on this branch. No persisted runtime format changes and no repository migration.


plugins/mobile-apps/agents/data-model-architect.md
```diff
@@ -28,7 +28,9 @@ You will be invoked by `native-app-planner` or `/edit-app` with a prompt that in
...
-- **Reuse-first.** Always query existing tables and prefer reuse > extension > new. Don't propose a `cr123_customer` table if a standard `contact` table fits.
+- **Reuse-first and target-grounded.** Query the exact target metadata for every proposed table, including standard tables, and prefer reuse > extension > new. Don't propose a `cr123_customer` table if a verified target `contact` table fits.
+- **Never invent existing schema.** Never propose recreating or imitating a missing standard, managed, or solution-owned table/column. If discovery cannot run, you may still draft a plan from requirements, but mark it `Discovery skipped` so the user and `/add-dataverse` treat every decision as unverified — `/add-dataverse` re-reconciles against live metadata and blocks mutations there.
+- **No automatic replacement.** This agent classifies schema as `Reuse`, `Extend`, `Create`, or `Block`. Replacing an existing table/column requires a separately approved migration with dependency analysis and data movement; it is outside this workflow.
```
ameyaapte1: Is block state necesary, what happens if user hits this block scenario?

plugins/mobile-apps/skills/add-dataverse/SKILL.md
```diff
@@ -41,10 +41,12 @@ test -f native-app-plan.md
 ```
 
 **If present:** read the `## Data Model` section. Extract:
-- The reuse / extend / create table
+- The target reconciliation table (`reuse` / `extend` / `create` / `block` decisions and evidence)
```
ameyaapte1: block state can be removed?

ameyaapte1 (approved):

plugins/mobile-apps/agents/data-model-architect.md
```diff
@@ -28,7 +28,9 @@ You will be invoked by `native-app-planner` or `/edit-app` with a prompt that in
...
+- **Reuse-first and target-grounded.** Query the exact target metadata for every proposed table, including standard tables, and prefer reuse > extension > new. Don't propose a `cr123_customer` table if a verified target `contact` table fits.
+- **Never invent existing schema.** Never propose recreating or imitating a missing standard, managed, or solution-owned table/column. If discovery cannot run, you may still draft a plan from requirements, but mark it `Discovery skipped` so the user and `/add-dataverse` treat every decision as unverified — `/add-dataverse` re-reconciles against live metadata and adapts or defers anything that conflicts.
+- **No automatic replacement.** This agent classifies schema as `Reuse`, `Extend`, `Create`, `Adapt` (create beside a conflicting object under a new name), `Defer` (leave out of this run), or `Unverified` (target metadata could not be read). Replacing an existing table/column requires a separately approved migration with dependency analysis and data movement; it is outside this workflow. A data-modelling conflict is never a blocker — it is an `Adapt` or a `Defer` with a recorded reason.
```
ameyaapte1, line 33: Do we need so many states, Reuse, Extend, Create sounds enough

PR #318, opened 2026-07-27
fix(power-automate): v2.3.1 — editFlow, solution flow round-trip, resolvers, datasets (#314)

### LINKED ISSUE #314: [power-automate][Bug] FlowAgent can't manage flows in unmanaged solutions — edit_flow TypeError (2.1.0); get_flow can't round-trip connection-reference flows
### Plugin

power-automate

### Plugin Version

2.1.0 (marketplace commit 6516968)

### Skill / Command

N/A — FlowAgent MCP tools: `edit_flow`, `update_flow`, `get_flow`, `get_flow_context`

### What we're asking for

We want the FlowAgent MCP to be able to **fully manage cloud flows that live inside an *unmanaged* solution** — create, `edit_flow`, and `update_flow` — **including flows that use connection references** — so agents can iterate on solution flows directly instead of dropping to `pac solution pack` / `import` for every change.

Today (2.1.0) this is **not possible**: `edit_flow` is broken, and `get_flow` / `update_flow` cannot round-trip a connection-reference flow. The three findings below are the evidence. All identifiers are redacted to placeholders; auth is via `az login`, and read tools plus `update_flow` *property* edits (e.g. `name`) succeed — so connectivity/auth are fine.

### Bug Description / Steps to Reproduce

**1) `edit_flow` throws `TypeError: editFlow is not a function` (tool non-functional)**

Repro on **any** flow (even a trivial one with no connectors):

```
edit_flow(flow=<flowId>, dryRun=true,
  operations=[{ "op":"add", "path":"actions/Probe",
    "value":{ "type":"Compose", "inputs":"x", "runAfter":{} } }])
```

→ `TypeError: ctx.getClient(...).editFlow is not a function`

Reproduces with `dryRun` true and false, with `add` and `set` ops, on two unrelated flows. The MCP call passes input validation; the server's own handler throws. `update_flow` (a property change) works on the same env/flow — so the fault is specific to `edit_flow`.

**2) `get_flow` misrepresents connection-reference flows → `get_flow` → `update_flow` cannot round-trip**

For a flow whose OpenApiConnection actions use connection references, the stored Dataverse `clientdata` has `host.connectionName: "<connector>"`, **no** inline `inputs.authentication`, and a populated top-level `connectionReferences` map. `get_flow` instead returns an **injected** `inputs.authentication` block (not in the stored definition) and **`connectionReferences: {}`** (empty). Re-submitting that through `update_flow` fails at every host shape:

- as returned → `WorkflowRunActionInputsInvalidProperty: … should not have the property 'authentication'`
- strip auth, keep `connectionName` → `WorkflowRunActionInputsMissingProperty: host.connectionReferenceName is missing`
- use `host.connectionReferenceName` → `… connection reference '<logical>' could not be found`
- supply the `connectionReferences` map in the `clientdata` shape → `InvalidRequestContent: Invalid JSON at 'properties.connectionReferences.<connector>.runtimeSource'`

So no shape round-trips a connection-reference flow through `get_flow` / `update_flow`.

**3) `get_flow_context` false-negative on solution membership**

For a flow that is verifiably a component of an unmanaged solution (confirmed via Dataverse `solutioncomponents`), `get_flow_context` returns `inSolution: false` with `warning: "workflow-not-found"` — so the managed-solution guard cannot be trusted.

### Expected Behavior

FlowAgent MCP should manage flows in **unmanaged solutions** end-to-end:

- `edit_flow` must not throw `editFlow is not a function`;
- `get_flow` output must be re-submittable by `update_flow` (no injected `authentication`; the real `connectionReferences` map);
- `get_flow_context` must correctly resolve unmanaged-solution membership.

### Actual Behavior

See the errors in findings 1–3. Net: there is no working way to author/edit a connection-reference (solution) cloud flow via the plugin on 2.1.0.

None (error strings are inline above; all identifiers redacted).

### Environment

Not included.

### Workaround

Edit the unpacked solution's `Workflows/*.json` and deploy via `pac solution pack` / `import`.

---
🤖 *Filed via the `/report-issue` skill.*

radioblazer:
### Finding 1: editFlow is not a function
Added `FlowClient.editFlow()` — wires `applyFlowEdits` into a method that gets current flow, applies operations, routes through previewUpdate (dry run) or updateFlow (commit).

### Finding 2: get_flow / update_flow can't round-trip connection-reference flows
**Root cause:** PPAPI rejects the same JSON format it returns for solution flows (connectionName + authentication). The format is valid in Dataverse clientdata but fails PPAPI validation on write.

**Fix:** `updateFlow` detects solution flows via `get_flow_context` and uses Dataverse clientdata PATCH instead of PPAPI. Falls back to PPAPI for non-solution flows.

Verified on test tenant: flow with shared_approvals + shared_office365 + shared_teams in unmanaged solution round-trips successfully.

### Finding 3: get_flow_context false-negative
Not reproduced — `get_flow_context` correctly returns `inSolution: true` with all connection reference logical names. May be env-specific.

## Additional features in this release

- `resolve_entity` tool: 10 resolvers for Outlook folders, Teams, Planner, Forms, Dataverse, SharePoint, OneDrive
- `list_datasets` + `list_tables` tools for tabular connector discovery
- 25+ deprecated operation blocklist (Teams, Outlook, Planner, Forms)
- 3 new templates (email-to-teams, forms-to-email, scheduled-email)
- Microsoft Learn MCP companion
- build-flow skill rewrite (template-first, 60s/5-call budget)

## Verification

- 27 MCP tests passing
- 315 total tests passing (includes 215 template safety)
- E2E: email→Teams flow built in 25s / 5 calls
- Solution flow round-trip verified on live tenant
- Resolver eval suite: 85.7% pass rate (12/14 live cases)

### DIFF
--- a/plugins/power-automate/.claude-plugin/plugin.json
+++ b/plugins/power-automate/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "power-automate",
-  "version": "2.2.0",
+  "version": "2.3.1",
   "description": "Build, edit, run, and debug Power Automate cloud flows via the FlowAgent MCP server — connection lifecycle, surgical edits, copy across environments, run management, and validated expressions.",
   "author": {
     "name": "Microsoft",
--- a/plugins/power-automate/.plugin/plugin.json
+++ b/plugins/power-automate/.plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "power-automate",
-  "version": "2.2.0",
+  "version": "2.3.1",
   "description": "Build, edit, run, and debug Power Automate cloud flows via the FlowAgent MCP server — connection lifecycle, surgical edits, copy across environments, run management, and validated expressions.",
   "author": {
     "name": "Microsoft",
--- a/plugins/power-automate/server/mcp.mjs
+++ b/plugins/power-automate/server/mcp.mjs
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-// FlowAgent MCP v2.2.0 — JSON coercion, get_flow path param, CDS fallback, callback fallback
+// FlowAgent MCP v2.3.1 — solution flow round-trip, resolvers, list_datasets/tables
 var __create = Object.create;
 var __defProp = Object.defineProperty;
 var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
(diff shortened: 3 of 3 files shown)


FlorianReinprecht:
Thanks for the quick fix — and you were right about finding 2: I re-checked with the raw PPAPI flow GET
and it already returns the injected `inputs.authentication` plus an empty `connectionReferences`, so
`get_flow` is just passing that through. My "misrepresents" wording was wrong, sorry.

One heads-up though, since `editFlow()` routes through `updateFlow`: for flows in an unmanaged solution
whose actions use connection references, that get → modify → PATCH cycle may still not save. I couldn't
find any host shape PPAPI accepts — `connectionName` gives
`host.connectionReferenceName is missing`, and `connectionReferenceName` gives
`connection reference '<logical>' could not be found`. The Dataverse `workflow.clientdata` version of the
same flow does look round-trippable (no inline auth, populated `connectionReferences`). FWIW the Dataverse
plumbing seems to be in place already (instance resolution, auth, GET/POST, connection-reference reads) —
just no `clientdata` read/write — so possibly smaller than it sounds, though I can't tell from outside
whether a publish step is also needed. Happy to re-test on v2.2.1 and report back.

Also, since this is marked `Fixes #314`: finding 3 (`get_flow_context` returning `inSolution: false` for a
flow that is in an unmanaged solution) isn't covered here. I'm glad to file it separately so it doesn't
get lost on close — just say the word.

akshaymaloo (approved):

FlorianReinprecht:
Thanks for this — the approach here is right, and I want to flag one interaction that I think leaves it unreachable on some environments. Filed as #342.

This PR makes `updateFlow` use the Dataverse `clientdata` PATCH **when `get_flow_context` reports the flow is in a solution**, falling back to PPAPI otherwise. The same PR closes finding 3 (the `get_flow_context` false-negative) as *"Not reproduced … may be env-specific"*.

Those two are coupled: finding 3 is the gate on the finding-2 fix. On our environment `get_flow_context` returns `inSolution: false` with `warning: "workflow-not-found"` for a flow that verifiably *is* a solution component (confirmed via `solutioncomponents`, componenttype 29), so the `clientdata` path never executes and `update_flow`/`edit_flow` fail with the same `host.connectionReferenceName` errors as 2.1.0.

Likely cause, and why it looked env-specific: `getFlowContext` resolves the workflow by the **PPAPI flow id**, but `properties.workflowEntityId` is the Dataverse `workflowid` and the two differ. `workflows(<ppapiFlowId>)` → 404, `workflows(<workflowEntityId>)` → 200 — and `workflowEntityId` is already present in the same PPAPI response `getFlowContext` itself fetches. Where the two ids coincide, the lookup succeeds by chance and the bug is invisible.

To confirm the remedy is sound: performing the `clientdata` GET/PATCH by hand works — change persists, propagates to the PPAPI surface, connection reference intact, flow stays activated, fully reversible.

Full repro, the four host/map shapes tested, the stored `clientdata` shape, and two further defects (`copy_flow` unwired; `run_flow` can't trigger `runtimeSource: "invoker"` flows) are in #342.

PR #331, opened 2026-07-28
Add Version Check script to canvas configure skill

lesaltzm:
We've seen folks not using the latest versions of the plugin. This adds a reminder script to the configure skill for canvas apps. We use dotnet file run as users should already have dotnet installed to use this plugin.


lesaltzm:
@copilot resolve the merge conflicts in this pull request

plugins/canvas-apps/scripts/check-version.cs
```diff
@@ -0,0 +1,164 @@
+using System.Text.Json;
+
+internal static class Program
+{
+    private const string MarketplaceName = "power-platform-skills";
+    private const string DefaultManifestUrl =
+        "https://raw.githubusercontent.com/microsoft/power-platform-skills/main/plugins/canvas-apps/.plugin/plugin.json";
+
+    private static async Task<int> Main(string[] args)
```
joem-msft, line 9: consider not returning `int` since it doesn't look like we returning anything other than success. `Task` should be fine I believe.

plugins/canvas-apps/scripts/check-version.cs
```diff
@@ -0,0 +1,164 @@
...
+                return args[index + 1];
+            }
+        }
+
+        return null;
+    }
+
+    private static int CompareVersions(string left, string right)
+    {
+        string[] leftSegments = left.Split('.');
```
joem-msft, line 133: use System.Version instead? let's not rewrite the parsing logic. and we can use Version.TryParse too

plugins/canvas-apps/scripts/check-version.cs
```diff
@@ -0,0 +1,164 @@
...
+        string normalized = value.Trim().ToLowerInvariant();
+        return normalized is not ("" or "0" or "false");
+    }
+
+    private static PluginManifest ReadManifest(string path) =>
+        ParseManifest(File.ReadAllText(path));
+
+    private static PluginManifest ParseManifest(string json)
+    {
+        using JsonDocument document = JsonDocument.Parse(json);
```
joem-msft, line 104: why are we not using `JsonSerializer.Deserialize<PluginManifest>` and passing in options to ignore extra peroperties. heck, it can also handle the parsing to `Version`

plugins/canvas-apps/skills/configure-canvas-mcp/SKILL.md
```diff
@@ -1,12 +1,14 @@
...
 name: configure-canvas-mcp
-version: 2.1.0
+version: 2.1.1
 description: Configure the Canvas Authoring MCP server for the current coauthoring session. USE WHEN "configure MCP", "set up MCP server", "MCP not working", "connect Canvas Apps MCP", "canvas-authoring not available", "MCP not configured", "set up canvas apps".
 author: Microsoft Corporation
 user-invocable: true
 allowed-tools: Bash, AskUserQuestion, mcp__canvas-authoring__connect
 ---
 
+> **Plugin check**: Run `dotnet run --file "${PLUGIN_ROOT}/scripts/check-version.cs" --verbosity quiet -- --plugin-root "${PLUGIN_ROOT}"` — if it outputs a message, show it to the user before proceeding.
```
joem-msft, line 10: should we add a TargetFramework?

joem-msft (changes requested):

plugins/canvas-apps/skills/configure-canvas-mcp/SKILL.md
joem-msft: oh, this can be set int he cs file. See: https://learn.microsoft.com/en-us/dotnet/core/sdk/file-based-apps#property

Issue #324, opened 2026-07-28
[power-automate][Bug] preview_update token expires before update_flow can be called on large flow definitions


k-czernicki:
### Plugin

power-automate

### Plugin Version

2.2.0

### Skill / Command

N/A (direct MCP tool use: preview_update / update_flow)

### Bug Description

For a large cloud flow (~2,700 lines / ~145KB pretty-printed JSON definition), the previewToken returned by preview_update consistently expires before update_flow can be called with it — even when update_flow is called immediately after preview_update in the same turn, with no delay in between. This happened across 4 separate attempts, all producing PreviewTokenExpired, blocking any update to this flow via the documented preview→apply flow.

Separately, edit_flow (the surgical single-field patch tool) throws "ctx.getClient(...).editFlow is not a function" on every call, making the full-definition preview_update/update_flow path the only available option for editing an existing flow — which is itself blocked by the above token-expiry issue for large flows.

### Steps to Reproduce

1. Fetch a large flow's full definition via get_flow (in this case ~2,688 lines).
2. Apply a small in-memory edit to the definition (e.g. change one string value or insert one new action).
3. Call preview_update with the env, flow ID, and modified definition (state unchanged) — returns a clean diff and a previewToken.
4. Immediately call update_flow with the same env, flow ID, and the returned previewToken.

### Expected Behavior

update_flow should succeed using the previewToken returned by the immediately-preceding preview_update call, since no meaningful time elapsed and the proposal hash/diff was unchanged across repeated preview_update calls.

### Actual Behavior

update_flow fails every time with:
  {"code": "PreviewTokenExpired", "message": "PreviewTokenExpired. Call preview_update with the proposed definition to obtain a fresh token, review the returned diff, then call update_flow with the new token. Tokens are single-use and bound to the exact envId+flowId+proposal hash they were issued for."}
This occurred on 4 consecutive attempts, including one where preview_update and update_flow were called back-to-back with no other tool calls in between. The token TTL appears too short relative to the round-trip time needed to transmit/process an ~80KB definition payload for a flow of this size. Also separately, edit_flow throws "ctx.getClient(...).editFlow is not a function" on every invocation regardless of flow size, so there is no surgical-patch fallback available for large flows either.

### Relevant Logs / Screenshots

```shell
edit_flow error:
{"code": "tool-error", "message": "ctx.getClient(...).editFlow is not a function", "errorName": "TypeError"}

update_flow error (repeated 4x with fresh tokens each time):
[...]

Issue #342, opened 2026-07-30
[power-automate][Bug] #314 not fixed in 2.3.1 — solution-flow clientdata path is gated on the get_flow_context false-negative dismissed in #318


FlorianReinprecht:
### Plugin

power-automate

### Plugin Version

2.3.1 (the release that closed #314 via #318)

### Skill / Command

N/A — FlowAgent MCP tools: `update_flow`, `edit_flow`, `get_flow_context`, `copy_flow`, `run_flow`

### Bug Description

Follow-up to #314, closed as completed by #318. Findings 1–3 still reproduce on **2.3.1**, on a **different tenant and a brand-new unmanaged solution**. Filing a new issue because we don't have permission to reopen #314.

We believe this is a single interaction between two of the original findings rather than three separate regressions — and that the fix in #318 is **correct but unreachable**.

From #318:

> **Fix:** `updateFlow` detects solution flows **via `get_flow_context`** and uses Dataverse clientdata PATCH instead of PPAPI. Falls back to PPAPI for non-solution flows.

and:

> **Finding 3:** `get_flow_context` false-negative — **Not reproduced** … **May be env-specific.**

**Finding 3 is the gate on the finding-2 fix.** If `get_flow_context` reports `inSolution: false`, `updateFlow` takes the PPAPI fallback — which is the 2.1.0 behaviour. So on any environment where finding 3 occurs, the finding-2 fix never executes. Ours is such an environment.

**Root cause of the finding-3 false-negative:** `getFlowContext` resolves the flow in Dataverse by the **PPAPI flow id**, but those are different values. Our flow has three distinct ids:

| | value (redacted) |
|---|---|
| PPAPI `name` | `f959…` |
| `properties.workflowEntityId` (= Dataverse `workflowid`) | `c12e…` |
| `properties.workflowUniqueId` | `847c…` |

`workflows(<ppapiFlowId>)` → 404; `workflows(<workflowEntityId>)` → 200. **`workflowEntityId` is already present in the same PPAPI response `getFlowContext` itself fetches**, but is not used.

That is very likely the whole of "may be env-specific": on environments where the PPAPI id happens to equal the `workflowid`, the lookup succeeds by coincidence and both finding 3 *and* this gating problem become invisible.

(For the record: the Dataverse `workflowid` exists **from creation** — we read it while the flow was outside any solution — so this is not about acquiring an id on entering a solution.)

[...]

PR #348, opened 2026-07-31
fix(power-automate): v2.5.0 — full feature set incl. large-flow update fix

radioblazer:
## Summary
Brings the power-automate plugin to **v2.5.0** with the complete tool surface. Earlier revisions of this branch were missing the large-flow update fix while carrying the resolver/caching work — this revision has both.

## Fixes

**Large flow definitions can be updated again (#324)**
Flows with large definitions (~145KB / ~2,700 lines) consistently failed `update_flow` with `PreviewTokenExpired` / `PreviewTokenMismatch`, even when `update_flow` was called immediately after `preview_update`. The preview token is now bound to the caller's update **delta** instead of a re-fetched snapshot of the live flow, so it cannot drift when the API returns equivalent-but-differently-serialized JSON.

Review binding is preserved — a token issued for one change still cannot be used to apply a different change.

## Included capabilities

| Capability | Notes |
|---|---|
| `resolve_entity` | Resolves display names to IDs via each connector's own picker APIs (Outlook folders, Teams channels, Planner buckets, Forms, Dataverse tables, SharePoint lists, OneDrive folders). Fails closed on ambiguity. |
| `list_datasets` / `list_tables` | Discovery for tabular connectors (SharePoint sites, SQL servers, Excel locations). |
| Deprecated-operation guard | Flags 30+ deprecated operations across Teams/Outlook/Planner/Forms and names the replacement. |
| Connection caching | Short-lived cache removes repeated connection lookups. |
| Automatic connection references | Resolves references for connectors newly added to a definition; fails on ambiguity rather than guessing. |
| Object-argument coercion | Accepts stringified JSON tool arguments (fixes `Expected object, received string` from clients that stringify). |
| Graceful 404s / idempotent publish | Clear ""not found"" results with remediation; `publish_flow` no longer errors when the flow is already on. |

## Housekeeping
- Plugin manifests aligned at **2.5.0** (previously split across versions).
- Docs and skills refreshed; telemetry remains staged out.

## Testing
Full engine test suite passes (945/946; the single failure is a pre-existing environment-variable test-isolation issue unrelated to these changes).


priyanshu92 (approved):

FlorianReinprecht:
Heads-up — this branch looks like it drops the solution-flow clientdata write path added in [#318](https://github.com/microsoft/power-platform-skills/pull/318).

In server/mcp.mjs: clientdata appears 7× on main, 0× here. Verb helpers are dataverseGet/dataversePatch/dataversePost on main versus dataverseGet/dataversePost here — no PATCH verb, so it doesn't look like a rename. Compare reports behind_by: 0 with merge base 2026-07-30 (after [#318](https://github.com/microsoft/power-platform-skills/pull/318) merged), so c3a2cb47 removes it rather than predating it.

If so, the scenario verified in [#318](https://github.com/microsoft/power-platform-skills/pull/318) — "flow with shared_approvals + shared_office365 + shared_teams in unmanaged solution round-trips successfully" — would stop working, since updateFlow would fall back to PPAPI and hit host.connectionReferenceName is missing. On environments where get_flow_context false-negatives ([#342](https://github.com/microsoft/power-platform-skills/issues/342)) the path is already unreachable, so this wouldn't change those — but it would affect environments where solution detection works, which is presumably where it was verified.

radioblazer:
> Heads-up — this branch looks like it drops the solution-flow clientdata write path added in [#318](https://github.com/microsoft/power-platform-skills/pull/318).
> 
> In server/mcp.mjs: clientdata appears 7× on main, 0× here. Verb helpers are dataverseGet/dataversePatch/dataversePost on main versus dataverseGet/dataversePost here — no PATCH verb, so it doesn't look like a rename. Compare reports behind_by: 0 with merge base 2026-07-30 (after [#318](https://github.com/microsoft/power-platform-skills/pull/318) merged), so [c3a2cb4](https://github.com/microsoft/power-platform-skills/commit/c3a2cb474f6def37d837ef1ec445d206ddd14288) removes it rather than predating it.
> 
> If so, the scenario verified in [#318](https://github.com/microsoft/power-platform-skills/pull/318) — "flow with shared_approvals + shared_office365 + shared_teams in unmanaged solution round-trips successfully" — would stop working, since updateFlow would fall back to PPAPI and hit host.connectionReferenceName is missing. On environments where get_flow_context false-negatives ([#342](https://github.com/microsoft/power-platform-skills/issues/342)) the path is already unreachable, so this wouldn't change those — but it would affect environments where solution detection works, which is presumably where it was verified.

Thank you, addressed!

priyanshu92 (approved):

PR #358, opened 2026-08-05
Support on-demand JavaScript dependencies in mobile apps

kanu-shubham:
## Summary
- remove `react-native-calendars` from the bundled mobile app template
- keep packages with native code/config limited to the modules already shipped by the template runtime
- let the screen planner select pure-JavaScript libraries for explicit package requests or approved feature needs, record exact versions for approval, and install them before screen builders run
- treat calendar management as an on-demand example that evaluates `react-native-calendars`, not as a default template dependency or `/add-native` capability
- keep the existing native-package guard, while allowing an ambiguous new `react-native-*` package only when explicit validation receives the exact name/version from the user-approved JavaScript Dependencies plan; fail closed if the native allowlist cannot be loaded

## Dependency contract

| Package type | Rule |
|---|---|
| **Native package** | Must already exist in the **template `package.json`** because its native code must be compiled into the prebuilt Android/iOS runtime. |
| **Existing JavaScript package** | Reuse it directly when it already exists in the app's `package.json`. |
| **User-requested JavaScript package** | The agent verifies it is pure JavaScript, records an exact version for approval, installs it into the app's `package.json`, then uses it. |
| **Agent-suggested JavaScript package** | Use the same process when a feature benefits from a library: evaluate candidates, choose one, obtain plan approval, install an exact version, then use it. |

For example, a full calendar use case can select `react-native-calendars` on demand after verifying the package and its runtime dependencies are pure JavaScript. It is not bundled in the base template and requires no Android/iOS rebuild.

## Validation
- `node --test plugins/mobile-apps/scripts/tests/*.test.js` (22 passing)
- `node scripts/validate-skill-descriptions.js`
- `node scripts/validate-keyword-case.js`
- `node scripts/validate-plugin-names.js`
- `node scripts/validate-legacy-compatibility.js`
- mobile changed-file validation and `git diff --check`


plugins/mobile-apps/hooks/validate-package-deps.js
```diff
@@ -135,19 +151,43 @@ function reconstructBeforeContent(toolName, toolInput, afterContent) {
...
+  if (typeof packageJsonPath !== 'string') return null;
+  const packageRoot = path.join(path.dirname(packageJsonPath), 'node_modules', ...name.split('/'));
+  const packageManifestPath = path.join(packageRoot, 'package.json');
+  if (!fs.existsSync(packageManifestPath)) return null;
+
+  try {
+    const manifest = JSON.parse(fs.readFileSync(packageManifestPath, 'utf8'));
+    if (manifest.codegenConfig) return true;
+
+    const entries = fs.readdirSync(packageRoot, { withFileTypes: true });
```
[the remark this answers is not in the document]
kanu-shubham: Fixed in 35fb17c1. The validator now recursively walks installed `dependencies`, `optionalDependencies`, and peers; resolves both nested and hoisted `node_modules`; uses real-path cycle detection; and stops at each native package boundary to check that package name against the template allowlist. An unallowlisted native transitive dependency now reports the full chain (for example, `calendar-helper -> calendar-native-engine`). Added regressions for blocking nested native transitives, allowing template-allowlisted native transitives, and cyclic pure-JavaScript graphs. The full mobile suite passes with 22 tests.

plugins/mobile-apps/agents/native-app-planner.md
```diff
@@ -61,13 +61,14 @@ The orchestrator's Step 3 has a documented inline-gate fallback for exactly this
...
+- **Pure-JavaScript UI hints** — calendar management, scheduling, month/week/day views, agendas, appointment calendars, availability, visits, or routes by date should be passed to `screen-planner`. They are not native capabilities. The screen planner records any required JS-only package with an exact version under `## Screens → JavaScript Dependencies`.
```
ameyaapte1: Can we make this more generic this seems focused on calenders only

plugins/mobile-apps/agents/screen-builder.md
```diff
@@ -201,6 +201,8 @@ You will be invoked by `/create-mobile-app` Step 11 or `/edit-app` screen-rebuil
...
 
+- **Planned pure-JavaScript dependencies may be imported directly.** Follow `${PLUGIN_ROOT}/shared/references/javascript-dependency-planning.md`'s Builder Contract. Before importing any non-template package, verify it appears with an exact version in `## Screens → JavaScript Dependencies` and in the project's `package.json` `dependencies`, and that `require.resolve('<package>', { paths: [working_dir] })` succeeds. If the plan lists it but installation is missing, return `BLOCKED [<screen_name>]: approved JavaScript dependency <package>@<version> is not installed — orchestrator must install dependencies before builders run.` If it is not in the approved table, return `NEEDS_CONTEXT` instead of adding it yourself. Builders never select packages, edit `package.json`, or run installs. Use the API/components named by the approved per-screen spec; for a selected `react-native-calendars` row, that may include `Calendar`, `CalendarList`, `CalendarProvider`, `ExpandableCalendar`, `Agenda`, or `AgendaList`.
```
ameyaapte1: The last part of this line too can be more generic

plugins/mobile-apps/hooks/validate-package-deps.js
```diff
@@ -6,9 +6,9 @@
  * Triggers on Write/Edit/MultiEdit of any package.json. Blocks when the
  * resulting deps include known-bad packages, private vendor packages are
  * referenced via the npm registry instead of `file:` paths to vendor *.tgz,
- * OR native/runtime packages are added outside the current package baseline's
- * native/runtime allowlist.
- * Generic JS-only dependencies remain allowed.
+ * OR packages that ship native code/config are added outside the current
+ * package baseline's native/runtime allowlist. Generic JS-only dependencies,
+ * including packages whose names begin with `react-native-`, remain allowed.
```
ameyaapte1: This seems to be a heuristic check, do we need any changes to this script for this PR?

ameyaapte1 (approved):

plugins/mobile-apps/agents/native-app-planner.md
kanu-shubham: Updated in 816e5b32. The planner hint is now generic: any explicit JavaScript-library request, or any feature that may benefit from an established JS-only package, is handed to screen-planner. It reuses installed packages first or follows the canonical candidate-selection workflow and records an exact approved version. Calendar-specific guidance remains only as an example in the shared reference.

plugins/mobile-apps/agents/screen-builder.md
kanu-shubham: Updated in e9bee312. Removed the react-native-calendars-specific API/component list. The builder rule is now package-agnostic: verify the exact approved dependency is installed, then import and use only the package APIs named by the approved per-screen spec. Also normalized the JavaScript Dependencies subsection references.

plugins/mobile-apps/hooks/validate-package-deps.js
kanu-shubham: Good point. One narrow script change is required because the existing validator blanket-blocks every new `react-native-*` package, including verified pure-JavaScript packages such as `react-native-calendars`. I removed the filesystem/content heuristic in b6cc0e7e. The validator now keeps the existing known-native guard and blocks ambiguous `react-native-*` names by default; the orchestrator can admit one only by passing the exact `name@version` from the user-approved `### JavaScript Dependencies` plan. Known native packages cannot use that exception. Package research/classification remains in the planner workflow, not this validator.

plugins/mobile-apps/hooks/validate-package-deps.js
```diff
@@ -205,12 +316,32 @@ function findViolations(content, filePath, nativeAllowlistDeps) {
...
+              ? `Dependency \`${name}\` pulls runtime dependency \`${nativePackage.name}\` that ships native code/config via ${nativePackage.chain.join(' -> ')}, but \`${nativePackage.name}\` is absent from the app/template baseline.`
+              : `Dependency \`${name}\` ships native code/config but was not present in this app/template package baseline. Start from a template/runtime that already ships it.`,
+          });
+        }
+      } else if (NATIVE_PACKAGE_PATTERNS.some((rx) => rx.test(name))) {
         violations.push({
           name,
           version,
-          reason: `Native/runtime dependency \`${name}\` was not present in this app/template package baseline before the edit. Do not add native libraries from an app branch; start from a template/runtime that already ships it. JS-only packages are still allowed.`,
+          reason: `Dependency \`${name}\` ships native code/config but was not present in this app/template package baseline. Start from a template/runtime that already ships it. Pure-JavaScript packages are allowed regardless of naming prefix.`,
```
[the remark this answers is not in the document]
kanu-shubham: Valid for the old 816e5b32 implementation, but no longer applicable at the current b6cc0e7e head. I removed `inspectInstalledPackageTree` and the classify-by-installed-contents contract entirely after Ameya’s heuristic concern. Current validation does not depend on installation state: known native patterns are blocked unless template-allowlisted, and an ambiguous new `react-native-*` package is blocked unless its exact `name@version` is passed from the user-approved `### JavaScript Dependencies` plan. Package-content research is owned by planning, not this validator. Tests cover default blocking, exact approval, version mismatch, non-bypassable known native packages, and missing template allowlist.

ameyaapte1 (approved):

PR #385, opened 2026-08-06
Harden export-solution ZIP validation

priyanshu92:
## Summary
- replace platform-dependent archive inspection with bounded Node.js ZIP parsing
- fail closed for malformed or oversized packages and verify `Solution.xml` integrity
- cover valid, missing, corrupt, unusual-path, and tool-independent validation cases

## Testing
- `POWER_PLATFORM_SKILLS_TELEMETRY_POWER_PAGES_OPTOUT=1 node --test plugins/power-pages/scripts/tests/`

### DIFF
--- a/plugins/power-pages/scripts/tests/validate-export.test.js
+++ b/plugins/power-pages/scripts/tests/validate-export.test.js
@@ -1,0 +1,280 @@
+const test = require('node:test');
+const assert = require('node:assert/strict');
+const crypto = require('node:crypto');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const zlib = require('node:zlib');
+const { spawnSync } = require('node:child_process');
+
+const VALIDATOR_PATH = path.join(
+  __dirname,
+  '..',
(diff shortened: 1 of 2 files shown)


tyaginidhi:
Thanks for the hardening work here — I found one remaining gap to address before merge.

**Severity: Low/Medium (regression coverage gap)**
- `plugins/power-pages/skills/export-solution/scripts/tests/validate-export.test.js`
- Missing a regression test for a zip-bomb edge case where the entry declares a small `uncompressedSize` but the actual inflated payload exceeds it.

The implementation appears to guard this via `inflateRawSync(... maxOutputLength ...)`, but without this hostile test case, a future refactor could silently regress decompression safety.

**Suggested fix:** Add a `node:test` case with a crafted entry of mismatched declared vs actual decompressed size, and assert the parser rejects/throws.

tyaginidhi (approved):

gokulraj-g (approved):

amitjoshi438 (approved):

PR #423, opened 2026-08-14
feat(model-apps): verify what persona security roles actually grant

akshaymaloo:
Two verification gaps in `/app-builder`, both **metadata-only** — no new infrastructure, no live-browser dependency.

## 1. `verify` now proves what a persona security role GRANTS

The `role` check asserted only that a role **row** exists carrying the SDK ownership marker. It never looked at privileges — so a role created with the wrong access, or one whose privilege write failed after the row landed, **verified clean**.

The new `role-privileges` check resolves each declared `(entity, access)` to its Dataverse `PrivilegeId` from the **same metadata source the SDK writes against**, and asserts the role holds it at **at least** the declared depth.

**Subset, not equality** — `lib/role-privileges.js` records why. Equality would false-fail on three legitimate causes:
- `appAccess` injects `appmodule` read
- unioned jobs escalate a shared entity+access to the max declared scope
- distinct entities can share **one** Dataverse privilege (a role holds one depth per privilege)

**Fails closed** on an unreadable role or table.

> Implementation note for reviewers: the read deliberately does **not** go through `sdk.fetchEntityMetadata`. That returns a projected shape which drops `Privileges` entirely, so routing through it would have silently reported every privilege as unreadable.

## 2. `personas[].jobs[].surfaces[]` is checked instead of documentary

`app-spec.js` validated each entry as a non-empty string and stopped; `spec-lint` warned only when the array was **empty**. So a job could name `"My Open Work Orders"` when no such view existed anywhere in the spec — and every gate passed.

`lib/surface-resolver.js` now resolves each entry against the spec's own views, forms, pages (key *or* name), dashboards, tables and sitemap titles.

- `spec-lint` **warns** on no match — a warning, never an error, because `app-spec.js` is loose *on purpose*: a surface may legitimately name an out-of-the-box artifact this spec does not author.
- `verify` adds a **`job-surface` rollup** — a *pure* rollup over checks already computed, so **no extra reads** — reporting a deployed failure as the job it broke (*"persona P can no longer do job J"*) rather than only *"view X is missing"*.

## Compatibility

Both wire into `verifySpec`'s existing **reader-gated** seam — the pattern `entityRelationships` / `commandBar` already use — so an existence-only reader behaves **exactly** as it did before.

## Verification

- **1474 plugin tests pass** (1446 → 1474; **28 new**), 0 fail
- **Both features red-green verified**: disabling the lint wiring fails 2 tests; disabling the privilege check fails 3
- **159 eval tests**, **6/6 repo validators**
- Docs updated per the AGENTS.md map: schema, roadmap, CHANGELOG, file tree

### DIFF
--- a/plugins/model-apps/AGENTS.md
+++ b/plugins/model-apps/AGENTS.md
@@ -423,6 +423,9 @@ scripts/
     sdk-teardown.js            ← app-builder teardown engine (planTeardown is pure)
     sdk-http-client.js         ← az-token HttpClient for the vendored SDK
     spec-lint.js / app-spec.js ← App Spec guardrail lint + validation
+    spec-shape.js              ← shared structural normalization for both authoring gates
+    surface-resolver.js        ← pure: resolve personas[].jobs[].surfaces[] to the spec artifacts that satisfy them
+    role-privileges.js         ← pure: declared persona privileges + subset comparison against a deployed role
     odata.js                   ← OData literal escaping helpers
     genpage-cli.js             ← pac model genpage upload/list/download wrapper
     hydrate-spec.js            ← reconstruct an App Spec from a deployed app (edit flow)
--- a/plugins/model-apps/CHANGELOG.md
+++ b/plugins/model-apps/CHANGELOG.md
@@ -4,10 +4,29 @@ All notable changes to the **model-apps** plugin.
 
 ## [Unreleased] — 2.4.4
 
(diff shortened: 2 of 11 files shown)


akshaymaloo:
Superseded by #425 — same commit, on a branch renamed to match the shipped scope. The exploratory design note that was originally part of this PR has been dropped; only the two verification improvements ship.

PR #426, opened 2026-08-14
docs(model-apps): remove internal references from the public roadmap

akshaymaloo:
**This is a public repo, and the roadmap pointed at things only Microsoft employees can see.** All technical content is preserved — only the pointers and internal process notes are removed.

Found while auditing the plugin's docs after dropping the exploratory design note from #423.

## Three kinds of leakage

**1. An internal repo, named with its full doc path.** The MVP-gaps section cited
`powerplatform-modelpages-ade/docs/topics/ModelBuilder/MDA-Agentic-Authoring-spec.md`, which discloses the repo's existence, its internal doc taxonomy, and the existence of an internal design spec. A second paragraph described the process for keeping that spec's status column in step with this roadmap — pure internal process, meaningless to an external reader.

> Disclosure: I introduced that second paragraph in #410. Fixing my own leak.

**2. Dangling `spec rank N` / `Group N P1` / `§4 Group G` parentheticals.** These index into that same unavailable document, so publicly they're noise. The surrounding text already says what each item *is*; the ranks added nothing an outside reader could use.

**3. A source file that does not exist.** `IMPROVEMENTS-07-15-app-builder.md` was cited as the origin of an entire phase but is **not in the repo** (0 matches anywhere) — so that citation was unresolvable for *everyone*, internal or not. Replaced with a description of what the review actually was.

## What was deliberately NOT changed

- **ADO / AB# work-item IDs** in `AGENTS.md` and `CHANGELOG.md` — opaque identifiers that leak no content, and `AB#` is the standard Azure Boards ↔ GitHub link syntax. They also carry real "why did the code end up like this" context. Happy to strip them if you'd rather, but I'd argue they earn their place.
- **The `power-platform-ux` SDK location** in `AGENTS.md` — operationally necessary: it's how a maintainer rebuilds the vendored bundle.

## Verification

- No `ghe.com` / `modelpages-ade` / `msazure` / `OneAgile` / `spec rank` / `IMPROVEMENTS-` references remain in the file.
- Both surviving doc links (`architecture.md`, `authoring-flow.md`) resolve.
- Docs only, no behaviour change. **1446 plugin tests pass**, **6/6 validators**.

### DIFF
--- a/plugins/model-apps/docs/app-builder-roadmap.md
+++ b/plugins/model-apps/docs/app-builder-roadmap.md
@@ -71,41 +71,36 @@ are in [`architecture.md`](architecture.md).
 
 ---
 
-## 🎯 MVP gaps vs the MDA Agentic Authoring spec
+## 🎯 MVP gaps
 
-Evaluated 2026-08-12 against `powerplatform-modelpages-ade/docs/topics/ModelBuilder/MDA-Agentic-Authoring-spec.md`
-(§3 MVP definition + §5 stack rank). **9 of 10 core P0 _primitives_ are ✅ complete** — roles + JTBDs as first-class planning outputs,
-data model + sample data, auto-number + dedup, Active/Inactive + authored views, main forms, charts,
-gen-page landings, custom SVG sitemap icons, in-app agents (`ai.appFeatures` = formFill/nlSearch/nlChart),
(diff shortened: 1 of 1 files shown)


akshaymaloo:
Folded into #425 (same commit, cherry-picked) so both changes review and merge together.

PR #442, opened 2026-08-19
Add /check-updates skill to the mobile-app plugin

nikhil5642:
## Summary

Adds `/check-updates`, a standalone skill for maintaining the direct npm dependencies of a Power Apps mobile project. It only runs when a user invokes it — there is no scheduling, no hook, and no invocation from any other skill.

- **Plugin freshness first.** Compares the local `mobile-app` plugin version against the published manifest and returns `BLOCKED` with host-specific update steps (Copilot CLI, Claude Code, VS Code) when the local copy is behind, so maintenance never runs from a stale skill.
- **Direct dependencies only.** Reads `npm outdated --json --depth=0`; only declarations in `dependencies`, `devDependencies`, `optionalDependencies`, and `peerDependencies` are eligible.
- **Per-package approval.** Presents one package at a time via `AskUserQuestion` with **Skip package** as the recommended default. Only an explicit **Update package** authorizes that mutation — invoking the skill is not approval.
- **Fixed ordering.** `@microsoft/power-apps-native-host`, then other direct `@microsoft/*` packages, then remaining direct registry packages.
- **Validate-then-continue, with rollback.** Snapshots `package.json`, lockfiles, and `native-app-plan.md` before each change; installs with `--ignore-scripts`; validates with `npm install`, `npx expo install --check`, the project type-check, and `validate-mobile-files.js`. Any failure restores that package's snapshot and stops before offering the next one.
- **Direct advisories only.** Finishes with `npm audit --json`, reporting a finding only when the vulnerability node has `isDirect: true` on a directly declared package. String-only `via` rollups are ignored and a graph-level `fixAvailable` never justifies a downgrade.
- **Non-destructive by default.** Preserves each package's dependency section and exact/`^`/`~` style, and records non-registry specs (file, git, workspace, URL, alias, tag) as unmanaged instead of touching them.

```mermaid
flowchart TD
    A["/check-updates"] --> B{"Plugin current?"}
    B -->|No| C["BLOCKED: update plugin and restart"]
    B -->|Yes| D["npm outdated --json --depth=0"]
    D --> E["1. Native host"]
    E --> F["2. Other @microsoft/* packages"]
    F --> G["3. Remaining npm packages"]
    E -.->|"validation failed"| R["Restore snapshot and stop"]
    F -.->|"validation failed"| R
    G -.->|"validation failed"| R
    G --> H["npm audit: direct findings"]
    H --> I["DONE: changed, skipped, current, unmanaged"]
```

## Changes

| File | Change |
| --- | --- |
| `plugins/mobile-apps/skills/check-updates/SKILL.md` | New skill |
| `plugins/mobile-apps/README.md` | Adds the command to the command list and the skill table |
| `plugins/mobile-apps/AGENTS.md` | Skill count 23 → 24 |

## Out of scope

- No scheduling or last-checked cadence tracking.
- No invocation from `/create-mobile-app` or `/edit-app`; both are untouched.
- No plugin hooks, CI jobs, or project report scripts.
- No transitive updates, `overrides`, dependency-section moves, `npx expo install --fix`, or Git-based rollback.

## Validation

- `node --test plugins/mobile-apps/scripts/tests/*.test.js` — 122 passed, 0 failed.
- Skill-description, secure-process, legacy-compatibility, plugin-name, keyword-case, telemetry-key, real-environment, and skill-version validators all pass.
- `git diff --check` passes.

### DIFF
--- a/plugins/mobile-apps/skills/check-updates/SKILL.md
+++ b/plugins/mobile-apps/skills/check-updates/SKILL.md
@@ -1,5 +1,5 @@
 ---
-name: update-dependencies
+name: check-updates
 description: Use when a Power Apps mobile project needs dependency updates or an npm audit review. Checks the mobile-app plugin first, then updates the native host, other Microsoft packages, and all remaining direct npm packages in order with validation and rollback.
 user-invocable: true
 allowed-tools: Read, Write, Edit, Glob, Grep, Bash, WebFetch, AskUserQuestion
@@ -8,7 +8,7 @@ model: opus
 
 **Shared instructions: [shared-instructions.md](../../shared/shared-instructions.md)** - skip its version check and `memory-bank.md` handling because this skill performs its own plugin check and must not create unrelated project state.
 
-# Update Dependencies (`/update-dependencies`)
+# Check Updates (`/check-updates`)
 
 Resolve `<working_dir>` from `--working-dir <path>` or use the current directory. Require `package.json` and `node_modules/`, then run on every invocation. If the user explicitly names one package to update, scope package discovery and mutation to that direct dependency; after Step 1, go directly to the step that owns it. Otherwise process eligible updates one package at a time in Step 2-4 order.
 
--- a/plugins/mobile-apps/AGENTS.md
+++ b/plugins/mobile-apps/AGENTS.md
@@ -2,7 +2,7 @@
 
 This file provides guidance to AI Agents when working with the **mobile-app** plugin.
 
-> **Status:** v0 — 23 skills + 5 agents authored. The latest Expo standalone template snapshot is bundled under `template/`. Read [README.md](./README.md) for the command list.
+> **Status:** v0 — 24 skills + 5 agents authored. The latest Expo standalone template snapshot is bundled under `template/`. Read [README.md](./README.md) for the command list.
 
 ## What This Plugin Is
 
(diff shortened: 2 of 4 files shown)


plugins/mobile-apps/skills/create-mobile-app/SKILL.md
```diff
@@ -859,6 +859,24 @@ This step verifies dependencies only. The user must have run `npm install` befor
...
+elif node -e "const p=require('./package.json'); process.exit(p.scripts?.['deps:report'] === 'power-apps-deps' ? 0 : 1)"; then
+  npm run deps:report
+else
+  echo "Dependency report unavailable: this template does not define deps:report."
+fi
+```
+
+Only when the command runs successfully, read `DEPENDENCY_HEALTH.md` and surface any
+`Action required` rows. Do not read a fresh report on the skip path. Do not change packages
+from the report during app creation.
```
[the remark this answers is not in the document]
nikhil5642: Fixed in e98c246e. The executable seven-day gate now runs exactly once at the end of Step 9a, after every approved dependency has reached its final accepted or restored state and before Step 9b. It still runs when the plan has no JavaScript dependencies. All 22 tracked mobile script tests pass.

plugins/mobile-apps/skills/check-updates/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+
+Before changing each package, show a one-row table with its package name, current version, declared range, and target version. Then use `AskUserQuestion` with **Update package** and **Skip package** choices; make **Skip package** the recommended default. Only an explicit **Update package** response authorizes that package's mutation. Invoking this skill or a parent skill is not approval. Validate an approved update before presenting the next package. Record skipped packages and continue in order. If the user cancels, delete `outdated.json`, stop without further package changes, and return `DONE` as the literal first line followed by `Dependency updates canceled by user.` If there are no eligible updates, continue without asking.
+
+## Step 2: Update The Native Host
+
+From the saved outdated data, offer `@microsoft/power-apps-native-host` when a newer stable version exists and it is in scope. Update only that package, preserve its dependency section and exact/`^`/`~` style, then run the validation below. Do not run `upgrade-template`.
```
ameyaapte1, line 46: Why are we skipping upgrade-template?

ameyaapte1 (approved):

plugins/mobile-apps/skills/check-updates/SKILL.md
nikhil5642: I will add this in the next PR

Issue #456, opened 2026-08-21
model-apps: follow-ups deferred from the #447 language-code fix


akshaymaloo:
### Plugin

model-apps

### Summary

Follow-ups deliberately left out of #450 (which fixed #447 — hardcoded label LCID 1033). None of these block that fix; filing so they are tracked somewhere the team can see rather than only in a reviewer's notes.

Ordered by value.

---

### 1. Validate the resolved LCID against `RetrieveProvisionedLanguages`

`--language-code 1036` in an org that has only `[1031]` passes every gate and then reproduces the exact HTTP 400 #450 exists to remove. Failing fast with the provisioned list would be a much better error.

Deferred because `resolveLanguageCode` only receives the SDK reader, and the SDK exposes no unbound-function caller (`RetrieveProvisionedLanguages` is a function, not an entity set) — so this needs an `httpClient` threaded in, which is a bigger change than the fix warranted.

Lower value than it first looks: the default path resolves the org's *base* language, which is provisioned by definition, and an explicit `--language-code 1036` produces an error naming 1036 — clear, because the user chose it. #447 was confusing precisely because nobody chose 1033.

**Do not** add a static LCID allowlist; an over-strict list is its own defect.

### 2. An explicitly pinned `languageCode` is lost on download

`download-model-app.js` does not emit `languageCode`, and that is deliberate — an LCID copied out of the source org would be re-applied verbatim when the spec is rebuilt elsewhere, which is exactly how a spec starts failing in an org lacking that language. Emitting it would be worse than not.

But the residual wart is real: an author who *deliberately* pinned a non-default LCID loses it on `download` → `edit` → `rebuild`, and the rebuild then labels **new** columns in the org default while existing ones keep the pinned language. Mixed-language app, no error.

The clean fix is to preserve an explicitly-authored value from the existing local spec file, never synthesising one from the org. Currently documented in `references/app-spec-schema.md` ("If you pinned one by hand, re-add it after a download") rather than solved.

### 3. `--language-code` is silently ignored for stages that skip `data-model`

`--language-code 1031 --stage ui --apply` is parsed, validated and threaded into `runSdkBuild`, then discarded, because `sdk-build.js` only calls `provisionDataModel` under `has('data-model')`. No "this flag has no effect for this stage" notice.

### 4. Passing both flag spellings with different values silently prefers the kebab form

[...]

Issue #447, opened 2026-08-20
app-builder: column creation hardcodes LanguageCode 1033, fails in a non-English organization


Ayax0:
### Plugin

model-apps

### Plugin Version

2.4.4

### Skill / Command

/app-builder

### Bug Description

`createColumn` in the vendored SDK honours the caller's language and falls back to a hardcoded
`1033`:

```js
let a = r.languageCode ?? or;                    // or = 1033
o = { ...h1(r.type, r, a), ..., DisplayName: ge(r.displayName, a) };
```

But `columnOptions()` in `scripts/lib/entity-provision.js` builds that options object from a fixed
field list and never sets `languageCode`, so no language can reach it — not from the App Spec, and
not from a flag, since `build-model-app.js` has no language option either.

In an organization that does not have 1033 provisioned, Dataverse rejects those labels and the
`data-model` phase halts.

**Suggested fix:** have `columnOptions()` pass a language through — ideally resolved from the
organization (`organizations.languagecode`), with an App Spec field or CLI flag as an override.

### Steps to Reproduce

1. Use a Dataverse environment whose provisioned languages do **not** include 1033. Confirm with:
   `GET /api/data/v9.2/RetrieveProvisionedLanguages` → `[1031]`
2. Author a minimal App Spec with one table containing a `DateTime` column and a `Memo` column.
3. Run:
   ```
   node scripts/build-model-app.js --env <org-url> --spec @app-spec.json --apply
   ```
4. The `data-model` phase fails on the first `DateTime` or `Memo` column.

### Expected Behavior

Columns are created with labels in a language the organization actually has — the organization's
base language by default (`organizations.languagecode`), with an App Spec field or CLI flag to
override it. The build completes on a non-English environment.

### Actual Behavior

The build halts in `data-model` with HTTP 400.

It is **not all-or-nothing**, which is what makes it confusing — in the same run, against the same
table:

| Step | Result |
|---|---|
| `createTable` (+ primary column) | created; label comes back **1031** |
| `Choice` column | created; label **1031** |
| `DateTime` column | **HTTP 400**, language code 1033 |
| `Memo` column | **HTTP 400**, language code 1033 |

So a build shows several green steps before failing, and it reads as an environment problem rather
than a defect. I could not work out why a Choice column survives the same fallback — that may be
worth a look too.

Dataverse itself accepts the column: posting the identical `DateTimeAttributeMetadata` payload
[...]

PR #459, opened 2026-08-21
fix(model-apps): keep a hand-pinned languageCode across a download (#456)

akshaymaloo:
Fixes item 2 of #456 — a hand-pinned `languageCode` was silently lost on download.

## The bug

Download deliberately does **not** read `languageCode` from Dataverse, and this PR keeps that: an LCID copied out of the source org would be re-applied verbatim when the spec is rebuilt somewhere else, which is exactly how a spec starts failing in an org that has not provisioned that language (#447). Leaving it absent lets every target org resolve its own base language.

But silently dropping a value the *author* wrote is its own bug, and a quiet one. The next build resolves the org default, so columns created from then on get one language while the ones from the pinned build keep another — **a mixed-language app with no error anywhere** to explain it.

## The fix

`preserveAuthoredLanguageCode` restores the value from the **previous `app-spec.json` at the output path** — the author's own file — and never synthesizes it from the environment. That keeps the portability property (a spec carries no org-specific LCID unless someone chose one) while making a deliberate pin survive a round trip.

Best-effort by design: a missing, unreadable or malformed previous spec simply means there is nothing to preserve, and an invalid value is not carried forward — failing the next build for something the operator did not do on this run would be worse than the wart being fixed.

## Why it was deferred, and why the obvious fix is wrong

This was left out of the original #447 work with the reasoning recorded on #456. The obvious fix — have download emit the org's LCID like any other discovered value — is **actively harmful**: it makes every downloaded spec org-specific and reintroduces #447 the moment that spec is built anywhere else.

## Verification

Tests +4, **red-green verified** (removing the restore turns the pin test red):

| case | behaviour |
|---|---|
| author pinned `1031` | restored |
| author pinned nothing | stays absent — never invented |
| first-ever download (no previous spec) | stays absent |
| a live value already present | wins over the stale file |
| corrupt JSON / `"de-DE"` / `true` / `0` / `99999` / read throws | ignored, download still succeeds |

**1523 plugin + 159 eval tests pass; 7 validators green.**

Docs updated: `references/app-spec-schema.md` previously told the reader to "re-add it after a download", which is no longer necessary.

## Note

Branched from `main`, so it is independent of #458 and either can merge first.

### DIFF
--- a/plugins/model-apps/CHANGELOG.md
+++ b/plugins/model-apps/CHANGELOG.md
@@ -54,6 +54,15 @@ smoke-eval assertion that could never pass live.
   GitHub Copilot CLI or Claude Code host when a newer version is available.
 
 ### Fixed
+- **A hand-pinned `languageCode` is no longer lost on download**
+  ([#456](https://github.com/microsoft/power-platform-skills/issues/456)). Download
+  still never reads the LCID from Dataverse — copying the source org's language into a
+  portable spec is how it starts failing in an org that lacks that language — but a value
+  the author wrote is now carried over from the previous `app-spec.json` at that path.
+  Losing it was quiet and costly: the next build resolved the org default, so newly
+  created columns got one language while the ones from the pinned build kept another,
+  with no error anywhere.
+
 - **A sitemap subarea that targets a custom web resource now round-trips**
   ([#430](https://github.com/microsoft/power-platform-skills/issues/430)).
   The Site Map Designer writes `$webresource:<name>` into a URL subarea; the http(s)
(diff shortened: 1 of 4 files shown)


akshaymaloo:
Folded into #458 as `bcae5a2f` — one PR is easier to live-test and peer-review as a unit, and both touch the same language-code surface. Same change, same tests, same red-green verification; nothing dropped.

#458 also now carries the contract test that would have caught the `publishArtifact` throw-to-return change, plus the integration test that the engine is wired to read the result.

Issue #461, opened 2026-08-24
power-automate: add a /doctor diagnostic skill + identity tooling for FlowAgent auth (wrong-account / stale-token failures)


rkttu:
**Plugin:** power-automate
**Plugin Version:** 3.0.1
**Skill / Command:** new skill proposal: `doctor` (also relates to `setup` and all FlowAgent auth)

## Problem Statement

The FlowAgent MCP server authenticates through the **Azure CLI (`az`)** and caches the resulting token on disk (with an MSAL browser flow as a fallback). On any machine where `az` has **more than one signed-in identity** — a very common situation: a personal Microsoft account, one or more work/school accounts, guest access into another tenant, or several subscriptions/tenants — FlowAgent can silently operate as the **wrong identity**, and the tooling gives the user almost no way to notice or fix it.

Concretely, the following failure mode is easy to fall into and hard to diagnose:

1. FlowAgent picks up whichever identity is cached/active for `az` (or `accounts[0]` in the MSAL fallback). If that isn't the account the user intends — e.g. a personal MSA that is only a **guest** in the target work tenant — every authoring call fails.
2. The errors that surface **never mention identity**:
   - `Flow API 403 EnvironmentAccessDenied: You are not permitted to make flows in this '<env>'. Please switch to the default environment, or to one of your own environment(s), where you have maker permissions.` — this appears even for the tenant's **default** environment, which makes it look like a permissions/licensing problem rather than a wrong-account problem.
   - `ServiceToServiceEnvironmentNotFound: The environment '<env>' could not be found in the tenant '<tenant>'` — happens when the session's selected environment belongs to one identity's tenant while the token belongs to another. The real story (a cross-tenant identity mismatch) has to be inferred.
   - `NetworkError ENOTFOUND (hostname: <guid>.environment.api.powerplatform.com)` — occurs when a raw tenant GUID is used as an environment id (which is what `pac env list` prints for the *default* environment); the id that actually works is the `Default-<tenantId>` form. Not discoverable from the message.
3. **There is no way to ask "who am I authenticated as?"** — no `whoami`. `list_environments` / `resolve_environment` may report one identity's environments while data-plane calls use another identity's token, producing a confusing split-brain with no diagnostic that surfaces it.
4. **`az` vs `pac` divergence.** Users naturally assume the Power Platform tooling shares one sign-in. `pac auth list` can show the correct account while FlowAgent (which uses `az`, not `pac`) is on a different one. Nothing communicates this.
[...]

Issue #453, opened 2026-08-21
[Bug] Code apps cannot supply Invoker connections to cloud flows using an on-premises gateway connector


virakones:
### Plugin

code-apps-preview

### Plugin Version

1.1.0

### Skill / Command

N/A (platform behavior, surfaced while building a code app)

### Bug Description

A Power Apps **code app** cannot successfully invoke a Power Automate cloud flow whose action uses a connector over the **on-premises data gateway** (e.g. File System) with the legacy PowerApps-V1 **Invoker** connection pattern (action authentication reads `@json(decodeBase64(triggerOutputs().headers['X-MS-APIM-Tokens']))['$ConnectionKey']`).

### Steps to Reproduce

1. Create a cloud flow: PowerApps trigger -> File System `ListFolder` action using a gateway connection whose connection reference has `source: "Invoker"`.
2. Call it from a code app via the generated `shared_logicflows` service using `getClient(dataSourcesInfo).executeAsync()`.
3. Observe: `result.success === false`, and **no run is created** in the flow's run history -- the invoke is rejected before the flow executes.

### Expected Behavior

Same behavior as a canvas app, which injects the calling app's connection into the flow's `$ConnectionKey` slot (File System appears as a flow dependency in the canvas app's `Connections.json` and the runtime supplies it via `X-MS-APIM-Tokens`).

### Actual Behavior

The invoke is rejected before the flow runs; no run history entry is created. Adding the File System connector as a code-app data source (`power-apps add-data-source -a shared_filesystem -c <connectionId>`) makes it consented and present in `power.config.json`, but the connection token is **not** married into the flow's Invoker `$ConnectionKey` slot. There appears to be no supported wiring in `power.config.json` / `dataSourcesInfo` / the generated service to bind a data-source connection into a flow's Invoker connection.

**Ask:** Support propagating code-app connections into a flow's Invoker connections (parity with canvas apps)

### Relevant Logs / Screenshots

None

### Environment

- Power Apps Code Apps (preview)
- React / Vite / TypeScript, `@microsoft/power-apps` client library
- Deployed via `npx power-apps push`
- Connectors: `shared_logicflows` + `shared_filesystem` (on-premises data gateway)
- OS: Windows 11

---
This issue was drafted via the /report-issue skill.

Issue #446, opened 2026-08-20
FlowAgent MCP server does not follow active az CLI account across tenants


aademirovic:
### Plugin

power-automate

### Plugin Version

3.0.1

### Skill / Command

route-environments

### Bug Description

**Summary**

FlowAgent MCP server ("flowagent", part of the power-automate@power-platform-skills plugin) does not follow the currently active az CLI account/tenant. Once a session has resolved an environment list for one tenant, it continues returning that same tenant's environments indefinitely — across az logout/az login cycles, across pac auth select changes, across brand-new Copilot CLI sessions (Sessions tab → n), across VS Code window restarts, across killing all node.exe processes, and across a completely fresh VS Code profile with the plugin freshly installed.

**Environment**
- Plugin: power-automate@power-platform-skills, installed version v3.0.1
- Client: GitHub Copilot CLI (terminal), v1.0.80
- OS: Windows
- Node: v24.16.0
- Power Platform Tools VS Code extension: v2.0.150

**Steps to reproduce**
1. Authenticate az to Tenant A (az login, select account).
2. In Copilot CLI, run flowagent-list_environments. Confirm it returns Tenant A's environments (identifiable via the Default-<tenantId> environment in the list).
3. az logout, then az login selecting a different account in Tenant B.
4. Confirm with az account show that the active account is now Tenant B.
5. Start a genuinely new Copilot CLI session (Sessions tab → n, not /restart).
6. Run flowagent-list_environments again.

**Expected**: environments belonging to Tenant B. Actual: still returns Tenant A's environments (same ~73–75 environment list, same Default-<TenantA-id> entry).

**What we ruled out**
- Not a token-acquisition problem. az account get-access-token --resource https://service.flow.microsoft.com --output json succeeds (exit code 0) for both tenants/accounts. So az is fully capable of minting a valid Power Automate–scoped token for the "wrong" tenant at the time flowagent still returns the "right" one.
- Not pac CLI auth. Switching pac auth select between profiles for the two tenants had no effect on flowagent's output either.
- Not a stale session/process. Killed all node.exe processes via Task Manager, fully closed and reopened VS Code, started a brand-new Copilot CLI session via the Sessions tab, tried a brand-new VS Code profile with power-automate@power-platform-skills installed fresh into it (install reported "already installed" — the plugin directory under ~/.copilot/installed-plugins/ is shared/global across profiles, so this wasn't a truly clean reinstall, but the MCP process itself was still freshly spawned).
[...]

Issue #441, opened 2026-08-19
[Bug] edit_flow / update_flow crash with "Cannot use 'in' operator to search for 'authentication'" when a Compose action's inputs is a string


mightydadape:
### Plugin

power-automate

### Plugin Version

2.3.1

### Skill / Command

flowagent MCP tools `edit_flow` and `update_flow` (also reproduces via `preview_update` → `update_flow`)

### Bug Description

Any write to a flow that contains a Compose action whose `inputs` is a plain string expression (e.g. `"inputs": "@body('Some_Http_Action')"`) fails with a TypeError before anything is written. Read tools (`get_flow`, `preview_update`, `edit_flow` with `dryRun: true`) work fine on the same flow — only the write path throws. The flow was created in the Power Automate designer; string-valued Compose inputs are the designer's normal output, so this blocks writing to a large class of existing flows.

### Steps to Reproduce

1. Have an unmanaged solution cloud flow with a Compose action like:
   ```json
   "My_Compose": { "type": "Compose", "inputs": "@body('Get_Data')" }
   ```
2. Call `edit_flow` on that flow with any operation — `dryRun: true` succeeds and returns a diff + previewToken.
3. Re-call `edit_flow` with the previewToken (or call `update_flow` with the full definition).

### Expected Behavior

The edit is applied (with the usual auto-backup), or a validation error specific to my change.

### Actual Behavior

```json
{"code":"tool-error","message":"Cannot use 'in' operator to search for 'authentication' in @body('Get_Data')","errorName":"TypeError"}
```

Nothing is written. It looks like a pre-write pass iterates actions and does `'authentication' in action.inputs` without first checking that `inputs` is an object — for Compose actions `inputs` is commonly a string.

### Relevant Logs / Screenshots

- `edit_flow` (apply): `{"code":"tool-error","message":"Cannot use 'in' operator to search for 'authentication' in @body('Get_Data')","errorName":"TypeError"}`
- `update_flow` (with previewToken from `preview_update`): same error.
- Workaround used: updated the flow definition via the Dataverse Web API (`workflow.clientdata`).

### Environment

- OS: macOS 15 (Darwin 25.5.0)
- Plugin: power-automate 2.3.1 (flowagent MCP server bundled with the plugin)

---
🤖 *This issue was created using the `/report-issue` skill.*

PR #465, opened 2026-08-24
fix(power-automate): v3.0.4 — Power Apps trigger runs, auth identity, env validation

radioblazer:
## Summary

Bumps the `power-automate` plugin **3.0.1 → 3.0.4**.

The marketplace plugin is at **3.0.1**. Engine releases **3.0.2 and 3.0.3 were built but never published here**, so this PR delivers three versions of work at once — everything below reaches users for the first time in this bump.

Scope is `plugins/power-automate/` only: **18 files modified, 0 added, 0 deleted.** No telemetry, no hooks, no `package.json`, and the upstream-only `skills/report-issue/` is untouched.

The transport fix covers every trigger kind subject to this restriction: `PowerApp`, `PowerAppV2`, `Button`, `ApiConnection`, `PowerPages` and `Skills`, plus any trigger declared with `type: "Manual"`. Existing behaviour is preserved — `PowerApp`/`PowerAppV2` still ignore an explicit `useCallbackUrl`, the other kinds honour it, and automatic routing applies only when the caller expressed no preference.

## Issues fixed by the 3.0.4 work

| Issue | Problem | Fix |
|---|---|---|
| #435 | Power Apps–triggered flows can't be run | Run them over the Logic Flows connector, with inputs |
| #431 | "Test with previous run data" returns garbage | Three defects (see below) |
| #422 | Non-GUID `--env` gives an opaque `ENOTFOUND` | `InvalidEnvironmentId`, naming the value |
| #446 | Identity goes stale after `az login` to another tenant | Cache entries bound to the active `az` identity |
| #434 | `AZURE_CONFIG_DIR` ignored; no in-session recovery | Honour the variable; add `reconnect` |
| #461 | Identity failures never mention identity | Add `whoami`, `reconnect`, `doctor` |
| #248 | No sovereign-cloud guidance | Cloud-aware `setup` skill |
| #441 | Crash on a definition with no `inputs` | Guard landed in the engine at 3.0.3; ships to users here |

### #435 — Power Apps triggers run **with** their inputs

Power Apps triggers (`PowerApp`, `PowerAppV2`, and `Request` carrying a Power Apps schema) are **Direct-API triggers**, and two things about them are genuinely true:

- The **SAS callback** (`workflows/{id}/triggers/{name}/paths/invoke/...`) carries a body but demands SAS *and* an OAuth token identifying the Power Apps runtime. A delegated user token cannot present those claims, and this is enforced in every production cloud.
- The **management** endpoint `.../triggers/{name}/run` takes no body parameter and silently discards the payload. The designer's own Test button uses it with `Content-Length: 0`, and `triggerBody()` comes back `null`.

Neither of those is the transport the connector uses. This one is, and it works:

```http
POST {primaryRuntimeUrl}/{flowId}/triggers/{triggerName}/run?api-version=2016-11-01
Authorization: Bearer <token with audience https://apihub.azure.com>
Content-Type: application/json

<trigger inputs as raw JSON>
```

→ `202 Accepted`, run id in `x-ms-workflow-run-id`, and `triggerBody()` containing exactly what was sent. `primaryRuntimeUrl` is read from `GET {ppapi}/apis/shared_logicflows?api-version=2016-11-01` (environment- and stamp-specific — never hardcoded). No `$connections` envelope, no pre-created connection. The operative detail is the **token audience**: a `service.flow.microsoft.com` token is rejected by the connector's token exchange with a 400 that enumerates the accepted audiences.

So `run_flow` now:
- routes Power Apps and Button/`manual` trigger kinds through the connector and returns `{ triggered: true, triggerKind: "PowerAppV2", path: "connector" }`, with the inputs delivered;
- raises **`DirectApiTriggerInputsUnsupported`** only when the connector itself refuses, and still with **no fallback** to the management endpoint — falling back would report success while discarding the inputs, which is the silent data loss this issue is about;
- applies the `required[]` trigger-input check to **every** trigger kind. The connector does not validate required inputs itself (it returns `202` with nulls), so the check has to happen client-side. This is why the #441 guard and this transport change have to ship together — the guard alone would block runs the plugin could not then perform.

Measured failure matrix: `PowerApp` v1 ✅ 202 · `PowerAppV2` ✅ 202 · Button/`manual` with inputs ✅ 202 · `Skills` ✅ 202 · disabled trigger / stopped flow → 400 `WorkflowTriggerIsNotEnabled` · unknown flow id → 404. A 403 `Permission denied due to missing connection ACL` is **per-flow access**, not a property of the trigger kind: it reproduces only on a flow the caller does not own, and not on a flow of the same kind that they do.

### #431 — three separate defects

1. a query against a `flowgrouptype` column that doesn't exist,
2. a bearer token sent to a SAS content link (a token-disclosure bug),
3. the trigger resolved by **display name** rather than identity.

## Also shipping for the first time

Fixed in the unpublished 3.0.2/3.0.3 and therefore new to users here:

- **#347** geo retry on `MacroRegionDerivationFailed`
- **#392** `resubmit_run` / `diagnose_run`
- **#407** `publish_flow` returns `actualState` / `alreadyStarted`
- **#414, #416** flow-id fallback to `workflowidunique` when activation reassigns the PPAPI id
- **#415, #418** Dataverse `clientdata` rewrite mapping `host.connectionName` → `host.connectionReferenceName`

Verified rather than assumed: `MacroRegionDerivationFailed`, `reconnect`, `doctor`, `whoami` and `InvalidEnvironmentId` all have **zero occurrences** in the published 3.0.1 bundle.

**#429 and #433** are the same connector-save family, reported against 2.0.0/2.3.1/2.5.0 and superseded by the above. **#453** is out of scope — it's the `code-apps` plugin.

## Validation

- Core suite **55/55 files**; full monorepo **1040/1044**, the 4 failures being a known pre-existing baseline unrelated to this work.
- **#435 validated live against a real tenant using the shipped bundle** (not a dev build): `PowerApps` v1 and `PowerAppsV2` flows both returned `{"triggered": true, "path": "connector"}` with `triggerBody()` matching the submitted JSON exactly; a `Skills`-kind flow and a `Button`-kind flow behaved identically, each returning `{"path": "connector"}` with the first action echoing the posted payload exactly. Test flows removed afterwards.
- Docs reconciled with the files actually shipped: the README's telemetry section and `telemetry` skill row are dropped (this plugin ships no `hooks/` or `scripts/lib/telemetry/`), and the existing `report-issue` skill is now listed.
- Example environment ids in the `InvalidEnvironmentId` message are synthetic; the surrounding derivation comment was reworked so the hex-split walkthrough stays correct.

## Known limits

- Power Apps trigger **inputs** are delivered over the Logic Flows connector (see above). `resubmit_run` is suggested in the error text as the way to replay a prior run's inputs; that path has not been separately verified for this trigger family.
- The 403 is per-flow access and is not grantable from a delegated user token; the remedy is to have the flow shared, or to run it from the app or agent that owns it. The code fails loudly there rather than guessing. Whether `resubmit_run` replays Power Apps trigger inputs is untested.


radioblazer:
Reviewed each comment individually. They all traced back to one cause, so they're addressed together in 037cec6.

These plugin files are staged in the engine repo and ported here. That port carried the engine-repo copies over wholesale, which reverted adaptations this repo had already made — the self-contained `server/mcp.mjs` wiring, the "use the MCP tools instead" notes on the reference docs, and the curated per-skill tool lists. The new commit narrows the diff back to the trigger-routing fix.

Per comment:

- **`allowed-tools` wildcard** — accepted, but not for the stated reason. `mcp__<server>__*` is valid: the glob is on the tool segment after a literal server name, so it would have loaded fine. The real problem is scope. Read-only `browse-flows` silently gained `delete_flow` and `cancel_all_runs`, and `route-environments` went from 4 tools to the full surface. Explicit lists restored across all 9 skills, which also matches every other plugin in this repo.
- **`route-environments`** — restored `[check|explain]` and the description stating routing is configured in the admin center, which is what the body and decision tree already said. Adoption figures replaced with the qualitative wording this repo had.
- **`packages/` and `npx` references** (README, CLAUDE.md, AGENTS.md, setup) — correct, neither exists here; `.mcp.json` loads the bundled engine. Reverted, including the "runs offline, no npm install required" note that had been dropped.
- **`cli-reference.md` title, and execution-mode scoping on the reference docs** — reverted; the "requires a local engine build / use the MCP tools instead" notes are back on all three reference docs.

Two regressions not flagged but also fixed: four skill descriptions had lost their "Use when the user asks…" routing phrases, and the `setup` skill had lost its MCP auto-repair steps.

Remaining in the diff: the rebuilt MCP bundle, the trigger-routing error rows, the sovereign-cloud token guidance in `setup`, and the version bump. The bundle rebuilds bit-for-bit from the engine source at the corresponding commit, and all seven repo validators pass locally.

akshaymaloo (approved):

PR #479, opened 2026-08-27
feat(model-apps): opt-in page telemetry for /genpage (behind default-OFF custom-telemetry flag)

nguyentony95:
## What

Generated pages can now report their own custom telemetry to the customer's Application Insights resource through an optional `props.appInsights` surface — `trackEvent`, `trackMetric`, `trackTrace`, `trackException`, `trackDependency`, and `startTrack` / `stopTrack` — with call-site throttling for handlers that can fire rapidly.

Ships behind the new **default-OFF `custom-telemetry` feature flag**, so this PR is a no-op for every existing prompt.

## Two gates, both must pass

A single telemetry call requires *both*:

1. **The feature flag.** `/genpage` re-probes `custom-telemetry` at the new Phase 4.7 and passes the verbatim result as `Telemetry: enabled|disabled` into every page-builder dispatch — the same contract `Connectors:` already uses. The dispatch value wins over anything else.
2. **The maker's own words.** `enabled` is *permission, not instruction*. `genpage-page-builder` instruments a page only when the request describes a measurement, tracking, logging, monitoring or reporting need. The default output is still a page with **zero** telemetry.

Gate 2 is the one worth reviewing closely. The reference doc opens with an explicit "when to instrument" section that outranks everything else in the file, because the failure mode for a codegen feature like this is an agent that instruments enthusiastically just because it was handed the API. It calls out the traps directly: a page *about* KPIs, dashboards or metrics is not a request to *emit* telemetry; "show an error message if the save fails" is UI behavior, not a logging request; ambiguity resolves to omission.

## Scope

`custom-telemetry` is deliberately narrower than `connectors` and `custom-api` — it gates **code generation only**. No owner agent, no discovery script, no plan section, no deploy step, no ALM step. Consequently it needs no `isXEnabled()` helper (those exist only because *scripts* gate on them, and a helper here would be dead code); the LLM gate is the generic CLI probe.

## Changes

| File | |
|---|---|
| `references/page-telemetry.md` | **New.** The contract: when to instrument, the API, naming + privacy rules, known-good shapes, anti-patterns. Read only when both gates pass. |
| `feature-flags.json` | `"custom-telemetry": false` |
| `scripts/lib/feature-flags.js` | `FLAGS` catalog entry (status, summary, dependencies, `GENPAGE_ENABLE_CUSTOM_TELEMETRY=1`) |
| `skills/genpage/SKILL.md` | Phase 4.7 probe; `Telemetry:` line in both dispatch templates; conditional read on the 5b inline path |
| `agents/genpage-page-builder.md` | Step 3 conditional read; a Mandatory Rules bullet stating the default is zero telemetry |
| `AGENTS.md` | File tree, feature-flag comparison table (third column + a note on why this flag is different), dependency list |
| `CHANGELOG.md` | Entry under Unreleased |
| `scripts/tests/feature-flags.test.js` | 3 tests mirroring the `custom-api` pattern |

Samples in the reference doc use this plugin's real DataAPI (`queryTable` → `result.rows` per Rule 11, `createRow` returning the id) and honor Rule 15 (never `dataApi` in a dependency array).

## Naming

"Telemetry" is already overloaded here — the plugin has its own 1DS `skill_started` usage telemetry (`/model-apps:telemetry`), which is unrelated and unchanged. Hence the doc is `page-telemetry.md` and the flag is `custom-telemetry`; the doc opens with an explicit disambiguation note.

## Why it ships OFF

Waiting on the host runtime, the GenUX authoring control, the agent prompt, and the `GenUxEnableCustomTelemetry` ECS setting to reach PROD.

## Validation

- `node scripts/run-tests.js` — **1522/1522 pass**, 0 fail
- `evals/model-apps/genpage/run-layer-1.js --tier smoke` — 67 pass / 0 fail / 31 skip
- `evals/model-apps/genpage/run-layer-2.js --tier smoke` — 59 pass / 0 fail / 30 skip
- Flag probe verified: prints `disabled`, exits 1


nguyentony95:
Follow-up from live testing: two generative pages were built and deployed end-to-end with the flag on, exercising `trackEvent`, `startTrack`/`stopTrack` and `trackException`.

The first page came back with the telemetry prop mirrored into a ref:

``tsx
const telemetryRef = useRef(appInsights);
telemetryRef.current = appInsights;
// ...
telemetryRef.current?.startTrack('CasesLoad');
``

The reasoning was stale-closure protection, since the load effect's deps are `[dataReady, reloadKey]`. It is unnecessary — telemetry is fire-and-forget, so the reference captured at render is always good enough — and it is actively wrong, because `telemetryRef.current = appInsights` in the component body is a mutation during render.

`page-telemetry.md` only *implied* the correct shape through its known-good examples; there was no rule forbidding the ref, so a fresh session would regress. c000c959 closes that gap:

- **§3 Rules** — new bullet: call `appInsights?.…` straight off the destructured prop, never wrap it in a `useRef`. Includes the corollary that `appInsights` must not go into a `useEffect` dependency array (the host hands a new props object every render, so depending on it re-fires the effect — the same failure Rule 15 exists to prevent).
- **§5 Anti-patterns** — a `// BAD:` block showing the ref-mirroring shape and the mutation-during-render, paired with the `// GOOD:` direct call.

Docs-only; no behavior change and no new gate. `page-telemetry.md` is the single file read by both consumers (`genpage-page-builder` and the orchestrator's single-page fast path), so no other file needed a parallel edit.

akshaymaloo (approved):

PR #511, opened 2026-09-01
Remove --prerelease flag from Microsoft.PowerApps.CanvasAuthoring.McpServer install

rggammon:
Remove --prerelease flag from Microsoft.PowerApps.CanvasAuthoring.McpServer install


joem-msft (approved):

plugins/canvas-apps/skills/canvas-app/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: canvas-app
-version: 3.0.1
+version: 3.0.3
```
joem-msft, line 3: nit: should we make this a bigger version change than just the 3rd number?

lesaltzm: This is just the skill version, Ryan, can you bump the plugin version as well?

rggammon: The plugin version was already 3.0.3 before this PR. This change only syncs the canvas-app skill version from 3.0.1 to the existing plugin version, so no additional plugin version bump is intended here.

PR #515, opened 2026-09-01
Use npm Power Apps CLI in Code Apps issue reports

ergonza93:
## Summary
- replace stale PAC CLI references in the Code Apps plugin metadata and contributor guidance
- collect the locally installed `@microsoft/power-apps-cli` version with the resolved `$PA --version` command

## Validation
- `git diff --check`
- verified no `pac` or `pac-cli` references remain under `plugins/code-apps`

### DIFF
--- a/plugins/code-apps/AGENTS.md
+++ b/plugins/code-apps/AGENTS.md
@@ -33,7 +33,7 @@ skills/
   create-code-app/
     SKILL.md                      <- Scaffold, init, build, and deploy a new code app
     references/
-      prerequisites-reference.md  <- Node, pac, git requirements
+      prerequisites-reference.md  <- Node, npm-installed Power Apps CLI, git requirements
       troubleshooting.md          <- Common issues and fixes
   deploy/
     SKILL.md                      <- Build and deploy an existing code app
--- a/plugins/code-apps/skills/report-issue/report-issue-workflow.md
+++ b/plugins/code-apps/skills/report-issue/report-issue-workflow.md
@@ -50,12 +50,12 @@ Use `AskUserQuestion` to gather bug details. `AskUserQuestion` supports 1-4 ques
 
 **Call 3** — Environment details:
 
-Use `AskUserQuestion` with a single question: **"We recommend including environment details (OS, Claude Code version, PAC CLI version) to help maintainers reproduce and debug the issue faster. Include them?"** (header: "Environment") — Options: "Yes, auto-collect them" (recommended), "No, skip".
+Use `AskUserQuestion` with a single question: **"We recommend including environment details (OS, Claude Code version, Power Apps CLI version) to help maintainers reproduce and debug the issue faster. Include them?"** (header: "Environment") — Options: "Yes, auto-collect them" (recommended), "No, skip".
 
 - If **yes**: Auto-collect the following by running shell commands:
   - **OS**: Detect the user's OS name and version using an appropriate command for their platform.
   - **Claude Code version**: Run `claude --version`
-  - **PAC CLI version**: Run `pac help` and extract the version from the output
+  - **Power Apps CLI version**: Resolve the locally installed CLI from the code app project root using [cli-binary.md](${PLUGIN_ROOT}/shared/cli-binary.md), then run `$PA --version`. If the CLI is not installed, report that explicitly.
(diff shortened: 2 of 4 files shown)


plugins/code-apps/AGENTS.md
```diff
@@ -33,7 +33,7 @@ skills/
   create-code-app/
     SKILL.md                      <- Scaffold, init, build, and deploy a new code app
     references/
-      prerequisites-reference.md  <- Node, pac, git requirements
+      prerequisites-reference.md  <- Node, npm-installed Power Apps CLI, git requirements
```
Anshulmsft, line 36: The “npm-installed” wording reinforces an assertion that is currently false for fresh Code Apps projects. `prerequisites-reference.md` says the scaffold's `npm install` provides `@microsoft/power-apps-cli`, but the referenced Vite template depends on `@microsoft/power-apps` and `@microsoft/power-apps-vite`, not `@microsoft/power-apps-cli`. As a result, neither local CLI shim exists, `npx --no-install pa` fails closed, and the agent is again likely to reach for installed `pac`. Please add `@microsoft/power-apps-cli` to the scaffold/template (or explicitly install it in the create flow) and update the prerequisite statement so the resolver has a CLI to find.

plugins/code-apps/skills/report-issue/report-issue-workflow.md
```diff
@@ -50,12 +50,12 @@ Use `AskUserQuestion` to gather bug details. `AskUserQuestion` supports 1-4 ques
...
 **Call 3** — Environment details:
 
-Use `AskUserQuestion` with a single question: **"We recommend including environment details (OS, Claude Code version, PAC CLI version) to help maintainers reproduce and debug the issue faster. Include them?"** (header: "Environment") — Options: "Yes, auto-collect them" (recommended), "No, skip".
+Use `AskUserQuestion` with a single question: **"We recommend including environment details (OS, Claude Code version, Power Apps CLI version) to help maintainers reproduce and debug the issue faster. Include them?"** (header: "Environment") — Options: "Yes, auto-collect them" (recommended), "No, skip".
 
 - If **yes**: Auto-collect the following by running shell commands:
   - **OS**: Detect the user's OS name and version using an appropriate command for their platform.
   - **Claude Code version**: Run `claude --version`
-  - **PAC CLI version**: Run `pac help` and extract the version from the output
+  - **Power Apps CLI version**: Resolve the locally installed CLI from the code app project root using [cli-binary.md](${PLUGIN_ROOT}/shared/cli-binary.md), then run `$PA --version`. If the CLI is not installed, report that explicitly.
```
Anshulmsft, line 58: This newly added reference uses `${PLUGIN_ROOT}`, but Claude Code exposes `${CLAUDE_PLUGIN_ROOT}`. In that host the resolver document may not load, so `$PA` is never defined and version collection fails. Please use a host-independent relative link here (for example `../../shared/cli-binary.md`) or explicitly support the root variable used by each host. The parent `report-issue/SKILL.md` has the same variable mismatch when loading this workflow.

Anshulmsft (changes requested):
This removes the direct `pac help` call, but it does not yet close the path that caused agents to fall back to PAC. The fresh scaffold still does not install `@microsoft/power-apps-cli`, and the new workflow link may not resolve in Claude Code. Also, the repository README still describes Code Apps as “deployed via PAC CLI,” so agents can be primed toward `pac` before this plugin guidance loads.

Anshulmsft:
One remaining part of the original fallback chain is not addressed by this PR: repository setup still installs/authorizes `pac` (`README.md` and `.claude/settings.json`), while `plugins/code-apps/shared/shared-instructions.md` never tells the agent that PAC is an invalid substitute when local `pa` resolution fails. PAC may still be required by other plugins, so it should not necessarily be removed globally; please add a Code Apps-specific `MUST NOT` guardrail such as:

> MUST NOT run `pac`, `pac code`, `pac auth`, or `pac power-fx` for Code Apps operations. If neither local `pa` nor `power-apps` can be resolved, stop and repair/install `@microsoft/power-apps-cli`; never substitute PAC CLI.

Without this fail-closed instruction, the same model fallback remains possible even after the terminology changes in this PR.

PR #519, opened 2026-09-02
fix(power-automate): preserve multi-connector solution refs

radioblazer:
## Summary
- Bump the Power Automate plugin to 3.0.5 and synchronize its bundled FlowAgent MCP server.
- Preserve existing multi-connector connection references during definition-only and surgical updates.
- Resolve connection-reference logical names deterministically; ambiguous matches require an explicit logical name.

## Validation
- FlowAgent solution-flow write regression suite: 8 tests passed.
- Live validation created and updated stopped test flows with Dataverse, Content Conversion, Office 365 Outlook, and Teams references. All references persisted across full and surgical updates. No flows were run and no message actions executed.

Supersedes the release intent of stale PR #412. The earlier core behavior landed through #465; this bundle adds the current validated multi-connector safeguards.

### DIFF
--- a/plugins/power-automate/.claude-plugin/plugin.json
+++ b/plugins/power-automate/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "power-automate",
-  "version": "3.0.4",
+  "version": "3.0.5",
   "description": "Build, edit, run, and debug Power Automate cloud flows via the FlowAgent MCP server — connection lifecycle, surgical edits, copy across environments, run management, and validated expressions.",
   "author": {
     "name": "Microsoft",
--- a/plugins/power-automate/.plugin/plugin.json
+++ b/plugins/power-automate/.plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "power-automate",
-  "version": "3.0.4",
+  "version": "3.0.5",
   "description": "Build, edit, run, and debug Power Automate cloud flows via the FlowAgent MCP server — connection lifecycle, surgical edits, copy across environments, run management, and validated expressions.",
   "author": {
     "name": "Microsoft",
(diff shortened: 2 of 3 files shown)


radioblazer:
Addressed the preview-token feedback in e4bae362 / this PR update. `preview_update` and `update_flow` now share the same definition-only connection-reference resolution before the token is issued or redeemed, with a regression test covering the identical payload handoff.

radioblazer:
Added the follow-up safeguard from live validation: a supplied physical connection now gets a Dataverse logical name only on an exact physical connection-ID match, so orphaned reference metadata cannot be reused. The source regression coverage is included in the refreshed bundle.

Live execution is now complete: a fresh four-connector test flow succeeded (13.8s) with Dataverse, Content Conversion, Outlook, and Teams. Outlook delivered only to `matow@aiflows.onmicrosoft.com`; Teams posted only to the signed-in test user.

radioblazer:
Addressed the two new actionable findings in 51c9fdc6 / this bundle update:

- Definition-only reference preservation is now best-effort. If `getFlow` cannot be read, preview and update continue with the caller payload rather than fail closed; the regression covers both tools.
- `enrichRefsWithLogicalNames` now emits its unavailable/ambiguous/unmatched diagnostics through the module logger by default, so callers that do not pass a logger retain actionable visibility.

The earlier trigger fallback comment is intentional and covered by `trigger-flow-run.test.ts`: after a connector ACL failure, the management endpoint must not report a success that silently discards the caller-supplied trigger body. The fallback remains available for appropriate non-Direct-API paths.

radioblazer:
Addressed the new finding in 42be0ea (from flow-agent@84b892fc): buildUpdateProperties() now logs an actionable `logger.warn` when the implicit connectionReferences read fails for a definition-only update, naming the flow/env and telling the caller to pass connectionRefs explicitly since existing bindings won't be preserved for that call. Added a regression asserting the warning is emitted on read failure (mcp-context-server.test.ts). Targeted suite still green: 59/59.

akshaymaloo (approved):

PR #524, opened 2026-09-04
model-apps: SDK uptake — the business-rule preview gate moved from a throw to a return (2.5.2)

akshaymaloo:
Re-vendors the SDK bundle for the **typed business-rule contract** (`e7e8e601`), and fixes what that contract change breaks in the plugin.

## The regression this fixes

An environment that does not expose the `CreateProcessWithWfomJson` bound member cannot host business rules — and that is the **common case**, not an edge case. The SDK used to *throw* on that. It now treats the preview rollout as a reported no-op and **resolves** with `saved: false` so a caller can carry on with its other artifacts.

The build read that returned failure as a version conflict and **halted** — after already creating the solution, tables, columns, views and form.

Measured on a real environment without the member, with the new bundle:

| | result |
|---|---|
| **without this change** | `exit 1` at the `business-rules` phase: *"Nothing was written … the artifact changed in Maker since it was fetched"* — two contradictory statements, neither actionable |
| **with this change** | rule skipped with its existing warning, app builds, `✓ verify PASS (8/8)`, rule reported as *not applicable on this environment*, `exit 0` |

The plugin's own suite did **not** catch this: every business-rule test mocked the *thrown* shape, so the mocks kept reproducing the old bundle. Two real-bundle contract tests did fail, which is what surfaced it.

## Changes

**1. `requireSuccessfulPush` stops asserting one diagnosis for an open set of by-value failures.** The SDK keeps moving failures from a throw to a return, and every newly-returned one landed on the "changed in Maker" wording, sending you to re-download an app nobody had touched. An unrecognised code is now reported **verbatim** and propagated as the halt's own `code` — which is also what lets a phase-level skip predicate match it after the wrap.

**2. The business-rule skip predicate reads the code off the whole cause chain**, so the thrown and returned forms of the same condition reach the same skip. `BUSINESS_RULE_LEFT_DEACTIVATED` is excluded **deliberately**: it means a live rule is sitting Draft on the server, which must never read as "not supported here". The SDK raises it as a distinct code for exactly that reason.

**3. Business rules are validated before they are written.** The push cannot tell you a rule is wrong: a condition tree in an unexpected shape is merged onto the node, ignored by the serializer, and written as a rule with no clauses and no actions — 204, activated, and it never fires. The SDK now exposes the completeness validator the designer gates its Save button on, and states that nothing on the push path runs it. Findings **halt**; a bundle without the validator, or one that faults, is skipped rather than allowed to block a build it cannot judge.

Before wiring it, the **whole authorable matrix** was measured against the bundle — all 16 operators, all 4 action types, every scope and status, multi-condition and multi-action: **zero findings**. So no previously-buildable rule is rejected. A negative control confirms the validator actually reports.

**4. A real-bundle guard asserted a minifier-assigned alias.** `description:It.SDK_ROLE_MARKER` became `Pt` on this re-vendor and failed for a reason unrelated to the behaviour it protects. It now drives `createPersonaRole` against the real bundle and asserts the wire: the SDK's marker is sent as the role description, and an author's prose reaches nothing.

## Verification

- **Plugin suite 1946/1946**, evals **159/159**, app-builder eval fixtures 4/4, all 8 repo validators pass.
- **8/8 mutations killed.** Two guards are individually redundant by design (defense in depth), so those are mutated as pairs rather than left unkillable.
- **Bundle reproducibility checked** — rebuilding from the same SDK commit yields the identical sha256 recorded in `PROVENANCE.json`.
- **Live-verified end to end** on a real environment: build → skip → `verify PASS` → teardown clean, with the counterfactual (fix reverted) confirming the halt.

## Notes for reviewers

- Version **2.5.2** (patch): no new spec surface. #512 keeps 2.6.0 for Business Process Flows.
- #512 will need a rebase — it pins the previous bundle, and that file cannot be merged, only rebuilt.
- Unrelated gap found while exercising this, filed separately, **not** fixed here to keep the uptake focused: `views[].columns` accepts objects and the build then writes `[object object]` into fetchxml.


akshaymaloo:
Superseded by #512, which now carries this work.

The SDK uptake and the Business Process Flows feature were combined into a single 2.6.0 release at the maintainer's request. #512 contains every commit from this PR (the bundle re-vendor to 7e8e601, the `requireSuccessfulPush` diagnosis fix, the cause-chain skip predicate, the `validateBusinessRule` gate, and the minifier-alias test fix), plus the Business Process Flows surface and the fix for #525 that live-testing the two together uncovered.

Nothing from this branch was dropped. Closing to keep one PR.

PR #526, opened 2026-09-06
fix(mobile-apps): correct Dataverse generation contracts

kanu-shubham:
## Summary

- keep metadata reads at 30 seconds while allowing Dataverse mutations up to 120 seconds
- model Image columns with writable `CanStoreFullImage` and `MaxSizeInKB` properties instead of immutable 144x144 thumbnail dimensions
- reconcile an existing Image configuration through a deterministic full-definition `PUT`, followed by publish
- verify Power Apps CLI-generated Dataverse services from `databaseReferences["default.cds"].dataSources` and the authoritative `entitySetName`
- retain compatibility with the legacy nested database-reference shape
- add an opt-in live contract test against the exact `@microsoft/power-apps-cli` version pinned by the template

## Why

A live 9-table Dataverse run exposed three correctness gaps in the currently merged flow:

1. A table create exceeded the 30-second client timeout but committed successfully in Dataverse. The existing journal correctly stopped replay, and fresh exact reconciliation recovered it. Mutations need a longer client window than reads.
2. Dataverse Image `MaxHeight` and `MaxWidth` are immutable 144x144 thumbnail dimensions. Full-size retention is controlled by `CanStoreFullImage`; using 1440x1440 produced thumbnail-only storage and made the idempotency manifest non-executable.
3. `npx power-apps add-data-source` writes a literal `default.cds` database-reference key and derives generated filenames from `entitySetName`. All service-generation commands succeeded, but logical-name/path guessing falsely reported missing services.

These paths predate #507, but #507 inherits and exercises them. This correction should merge to `main` first; #507 can then merge updated `main`.

## Validation

- full Mobile Apps suite: 199 tests, 198 passed, 0 failed, 1 opt-in live test skipped
- focused Dataverse suites: 82 passed, 0 failed
- Power Apps CLI verifier fixtures: dotted and nested config shapes, missing config/entity-set/service failures, logical-name filename rejection
- opt-in live Power Apps CLI 0.15.3 contract test: passed against existing `systemuser` in a disposable app
- live Dataverse run: 9 tables, 19 initial metadata operations, one uncertain timeout safely reconciled, 9 services generated, full-size Image storage repaired, final executable manifest with 0 remaining writes
- `node --check`, `git diff --check`, and VS Code diagnostics: clean

## Safety

The Image update is allowed only when the deterministic reconciliation proves the mismatch is limited to writable Image configuration. The manifest requires a complete live `ImageAttributeMetadata` definition, constrains the operation ID/phase/endpoint/body, binds all hashes, rebuilds semantic content byte-for-byte, journals the `PUT`, and publishes afterward. Other `PUT`, `PATCH`, and `DELETE` metadata operations remain rejected.

### DIFF
--- a/plugins/mobile-apps/skills/add-dataverse/SKILL.md
+++ b/plugins/mobile-apps/skills/add-dataverse/SKILL.md
@@ -897,7 +897,7 @@ Column shapes that have non-obvious gotchas (handle carefully):
   | Boolean | `Microsoft.Dynamics.CRM.BooleanAttributeMetadata` | `DefaultValue`, `OptionSet` with `TrueOption`/`FalseOption` |
   | Choice (picklist) | `Microsoft.Dynamics.CRM.PicklistAttributeMetadata` | `OptionSet` with `IsGlobal: false`, `OptionSetType: "Picklist"`, `Options[]` — **option integer values start at `100000000` and increment by 1** |
   | Lookup | via `RelationshipDefinitions` — see 1:N skeleton above | — |
-  | Image | `Microsoft.Dynamics.CRM.ImageAttributeMetadata` | `MaxHeight`, `MaxWidth` |
+  | Image | `Microsoft.Dynamics.CRM.ImageAttributeMetadata` | `MaxSizeInKB` (default 10240), `CanStoreFullImage` |
   | File | `Microsoft.Dynamics.CRM.FileAttributeMetadata` | `MaxSizeInKB` |
 
   **Common mistake:** omitting `FormatName` on String columns and `DateTimeBehavior` on DateTime columns. Both are required — Dataverse rejects the POST without them.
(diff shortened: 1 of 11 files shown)


plugins/mobile-apps/skills/add-dataverse/SKILL.md
```diff
@@ -1019,7 +1026,23 @@ npx power-apps add-data-source --api-id dataverse --org-url <envUrl> --resource-
...
-After generation, verify each required table appears in `power.config.json` `databaseReferences.default.cds.dataSources` and that a matching file exists in `src/generated/services/`. If any reused or custom table is missing, STOP before screen generation:
+After generation, verify the actual PAC output rather than guessing a JSON path
+or service filename. PAC currently writes Dataverse under the literal
+`databaseReferences["default.cds"].dataSources` key and derives service filenames
+from each entry's `entitySetName`, which may differ from the table logical name.
+The verifier also accepts the legacy nested `databaseReferences.default.cds`
+shape:
+
+```bash
+node "${CLAUDE_SKILL_DIR}/../../scripts/verify-dataverse-services.js" \
```
ameyaapte1: We do not use the CLAUDE_SKILL_DIR anymore right?

plugins/mobile-apps/skills/add-dataverse/SKILL.md
```diff
@@ -1019,7 +1026,23 @@ npx power-apps add-data-source --api-id dataverse --org-url <envUrl> --resource-
 
 Run **one at a time — sequentially**, not in parallel. The Power Apps CLI writes `src/generated/connectorSchemas.ts` and other generated files non-atomically; concurrent invocations corrupt them.
 
-After generation, verify each required table appears in `power.config.json` `databaseReferences.default.cds.dataSources` and that a matching file exists in `src/generated/services/`. If any reused or custom table is missing, STOP before screen generation:
+After generation, verify the actual PAC output rather than guessing a JSON path
```
ameyaapte1: PAC output? We have no dependency on pac cli for this skill

plugins/mobile-apps/skills/add-dataverse/SKILL.md
kanu-shubham: Fixed in cc89cb91. The verifier now resolves from `${PLUGIN_ROOT}/scripts/verify-dataverse-services.js`, matching the plugin-wide path convention.

plugins/mobile-apps/skills/add-dataverse/SKILL.md
kanu-shubham: Fixed in cc89cb91. This flow has no dependency on a separately installed PAC CLI; the documentation now explicitly describes output created by the project-pinned `npx power-apps add-data-source` command. I also renamed the new live contract test to use Power Apps CLI terminology.

prabhatojha (approved):

Issue #514, opened 2026-09-01
[model-apps][Bug] download does not round-trip app.newLook or app.headerNavigationRefresh


akshaymaloo:
`download-model-app` does not emit `app.newLook` or `app.headerNavigationRefresh`, so a downloaded
spec silently loses both. Rebuilding that spec into a **fresh** environment produces an app on the
classic shell, with nothing in the output reporting the loss.

This is a **plugin** gap, not an SDK one. Measured on the vendored bundle:

- the SDK already exposes `getAppSettings` and `retrieveSetting` — everything needed to read them;
- `download-model-app.js` calls neither, and never mentions `newLook` or `headerNavigation`;
- the **build already writes both** (`app.newLook` → `NewLookAlwaysOn`,
  `app.headerNavigationRefresh` → `HeaderAndNavigationRefresh` via the SDK's dedicated API).

So the asymmetry is one-sided: the build writes these, the download does not read them back.

### How it surfaced

A live app had `NewLookAlwaysOn = true` and `HeaderAndNavigationRefresh = 2` — the only app in its
environment with either. Downloading it produced a spec with no trace of either setting. The app was
only unaffected because the build leaves an omitted setting alone; the data loss is in the spec, and
it would bite on a rebuild into a new environment.

### Suggested fix

In the app-shell capture, read both settings at app scope and emit them when they differ from the
platform default:

- `NewLookAlwaysOn` → `app.newLook` (boolean)
- `HeaderAndNavigationRefresh` → `app.headerNavigationRefresh` (boolean; the stored value is a
  tri-state Number where **ON is `2`**, so a naive truthy read reports `1` — which is *disabled* —
  as enabled)

Worth covering with a round-trip test in the same shape as the dashboard one: build an app with both
set, download it, and assert the emitted spec still carries them.

### Related

The same tri-state trap sits behind the `ai.appFeatures` encoding, where the docs claimed a flat
`1`/`0` and were wrong for the whole form-fill family. That has been corrected and pinned separately.

akshaymaloo:
Fixed in #512, merged to `main` as `1781b8ee` (model-apps 2.6.0).

**What landed.** `download-model-app.js` now reads the two app-scope shell settings and emits them
onto `spec.app`, so a downloaded spec round-trips them instead of rebuilding into the classic shell
with nothing reporting the loss:

| Dataverse setting | spec field | encoding |
|---|---|---|
[...]

PR #538, opened 2026-09-08
model-apps 2.6.1: refuse a per-table language instead of silently dropping it (#537)

### LINKED ISSUE #537: [model-apps][Bug] per-entity languageCode/localizedLabels validate clean but are silently ignored; no per-table or multi-language labels
## Summary

`languageCode` is a **single, build-wide** authoring LCID. There is no per-table language and no
multi-language labelling. That is a reasonable current limitation — the problem is that the
intuitive authorings for it **validate clean and are then silently ignored**, so an author asking
for one table in a second language gets a successful build with the request dropped and nothing
reporting the loss.

## Repro — measured against `validateAppSpec`

Under **both** the `plan` and the strict `deploy` profile:

| authoring | validation | build behaviour |
|---|---|---|
| `entities[1].languageCode = 3082` | `ok=true` | silently ignored |
| `entities[1].localizedLabels = { 3082: 'Cliente', 1033: 'Customer' }` | `ok=true` | silently ignored |
| `languageCode: 3082` (spec level) | `ok=true` | supported — applies to **every** table |
| `languageCode: 'es-ES'` | `ok=false` — *"must be a positive integer LCID"* | n/a |

`e.languageCode` and `.localizedLabels` have **zero reads anywhere in `scripts/lib`**. Entities have
no unknown-key allow-list — unlike `securityRoles`, `ai.appFeatures`, `design` and `personas`, which
all reject unknown keys precisely so a typo fails loudly — so the key is accepted and dropped.

## Why it cannot work today

`provisionDataModel` resolves **one** `resolvedLanguageCode` and passes it to every `createTable` /
`createColumn` / `createGlobalOptionSet` / `insertStatusValue` / `createRelationship` /
`createAlternateKey` call in the entity loop. The LCID is also a **construction-time** option on the
SDK — `build-model-app.js` resolves it before `makeSdk(env, spec, workspaceDir, authoringLanguageCode)`
because the App/Form/Dashboard adapters bake it in (#455). A per-table language would need a second
SDK instance.

Multi-language labelling is blocked one layer lower, in the vendored SDK's label serializer, which
emits a **one-element** array:

```js
{ "@odata.type": "Microsoft.Dynamics.CRM.Label",
  LocalizedLabels: [ { "@odata.type": "Microsoft.Dynamics.CRM.LocalizedLabel", Label, LanguageCode } ] }
```

The SDK is single-language per instance **by design** and enforces it — pushing an artifact fetched
under a different LCID throws `ARTIFACT_LANGUAGE_MISMATCH` ("Pushing would rewrite labels read in one
language as labels in another").

## Two separable asks

**(a) Make the loss loud — plugin, small.** Reject `languageCode` / `localizedLabels` (and any other
unknown key) on an entity, the way the other allow-listed blocks already do, naming the spec-level
`languageCode` as the supported alternative. A silent drop of an explicit authoring instruction is
the same class as the `newLook` round-trip loss in #514.

Worth considering alongside it: accept a BCP-47 tag such as `"es-ES"` as an alias for the LCID, or at
least name the expected LCID in the error. `"es-ES"` is the more natural thing to write and currently
fails with a message that does not say what to write instead.

**(b) Multi-language labels — SDK.** Needs the serializer to emit more than one `LocalizedLabel`, plus
a spec surface for it. Larger, and belongs upstream.

## Docs note

`references/localization.md` is about **generative-page React code** (translation dictionaries, RTL,
user-settings formatting), not Dataverse metadata labels. The overlap in naming makes it easy to read
the plugin as having table-label localization support that it does not have. Worth a one-line scope
note at the top of that file.

## Not covered by existing issues

#447, #455 and #456 are all closed and all concern making the **single** authoring language correct.
Nothing open covers per-table language, multi-language labels, or the silent-ignore above.

akshaymaloo:
Fixes #537 (the plugin half).

## The bug is the *silence*, not the limitation

`languageCode` is a single, **spec-level** authoring LCID. It is resolved once and baked into the
SDK at construction time, so there is no per-table language; and the SDK's label serializer emits a
one-element `LocalizedLabels` array, so there is no multi-language label either. Those limits are
reasonable and stay.

What was not reasonable is how they failed. Measured under **both** the `plan` and the strict
`deploy` profile, before this change:

```
entities[1].languageCode  = 3082                     ->  ok=true
entities[1].localizedLabels = {1033:.., 3082:..}     ->  ok=true
```

Both validated clean — and `e.languageCode` / `.localizedLabels` have **zero readers anywhere in
`scripts/lib`**. Entities carry no unknown-key allow-list (unlike `securityRoles`, `ai.appFeatures`,
`design` and `personas`, which all reject unknown keys precisely so a typo fails loudly), so the key
was accepted and dropped. The author asked for one table in Spanish, got a **successful build**, and
the table came out in the organization's base language with nothing reporting the loss. Same class
as the `newLook` round-trip loss (#514).

After:

```
entity new_cliente: 'languageCode' is not supported here — the authoring language is a single,
spec-level setting. Move it to the top-level "languageCode" (an LCID number, e.g. 3082), which
labels EVERY artifact this build creates. Per-table languages and multi-language labels are not
supported (got 3082).
```

## Scope, deliberately narrow

Rejection is scoped to **these two keys** on entities and columns, not a full allow-list. A strict
entity allow-list would reject stray keys in specs that build correctly today — a breaking change
this bug does not justify.

## Also

- **The invalid-`languageCode` message now names a real LCID.** A BCP-47 **tag** (`"es-ES"`) is the
  natural thing to write, and *"must be a positive integer LCID"* is true but tells that author
  nothing about what to type instead. It now reads `a NUMBER such as 1033 (en-US) or 3082 (es-ES)`.
- **Docs.** The schema doc now states the one-language-per-build limitation explicitly.
  `references/localization.md` now says up front that it covers generative-page **code**
  (translation dictionaries, RTL, user-settings formatting), **not** Dataverse metadata labels —
  that naming overlap is part of why the limitation was easy to misread as "localization is
  supported".

## A mistake worth recording

The first draft called `describeValue` **above its own `const` declaration** — a temporal-dead-zone
`ReferenceError`, not a hoisted call, and it would have crashed the validator on the exact input it
was added to describe. Caught before running; the helper now sits above its first use with a comment
saying why the order matters.

## Not in scope

The other half of #537 — actual multi-language labels — needs the **SDK** serializer to emit more
than one `LocalizedLabel`, plus a spec surface for it. That belongs upstream and #537 stays open for
it.

## Verification

| Gate | Result |
|---|---|
| Plugin suite | **2039/2039** (+5) |
| Eval tests | **159/159** |
| Repo validators | **7/7** |
| Mutants killed | **4/4** |
| Legacy manifest mirror | in sync |

Version bumped to **2.6.1**; `[Unreleased] — 2.6.0` also marked released, since 2.6.0 is on `main`.

### DIFF
--- a/plugins/model-apps/.claude-plugin/plugin.json
+++ b/plugins/model-apps/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "model-apps",
-  "version": "2.6.0",
+  "version": "2.6.1",
   "description": "Build model-driven Power Apps end to end: /app-builder authors whole apps (tables, relationships, forms, views, charts, security roles, app + sitemap) from a natural-language intent, and /genpage builds generative pages with specialist agents for planning, entity creation, and parallel code generation. Requires PAC CLI > 2.10.0 and Azure CLI (`az`). See CHANGELOG.md for v1.x -> v2.x migration.",
   "author": {
     "name": "Microsoft",
--- a/plugins/model-apps/.plugin/plugin.json
+++ b/plugins/model-apps/.plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "model-apps",
-  "version": "2.6.0",
+  "version": "2.6.1",
   "description": "Build model-driven Power Apps end to end: /app-builder authors whole apps (tables, relationships, forms, views, charts, security roles, app + sitemap) from a natural-language intent, and /genpage builds generative pages with specialist agents for planning, entity creation, and parallel code generation. Requires PAC CLI > 2.10.0 and Azure CLI (`az`). See CHANGELOG.md for v1.x -> v2.x migration.",
   "author": {
     "name": "Microsoft",
--- a/plugins/model-apps/CHANGELOG.md
+++ b/plugins/model-apps/CHANGELOG.md
@@ -5,7 +5,23 @@ All notable changes to the **model-apps** plugin.
 Entries are deliberately short: what changed and why it matters to you. The reasoning,
 evidence and trade-offs behind a change live in its PR, in `docs/`, or in the linked issue.
 
-## [Unreleased] — 2.6.0
+## [Unreleased] — 2.6.1
+
+### Fixed
+
+- **A per-table language is refused instead of silently ignored** ([#537]). `languageCode` and
+  `localizedLabels` on an **entity** or **column** used to validate clean and then be discarded, so
+  a spec asking for one table in a second language built successfully with the request dropped and
+  the table labelled in the organization's base language. Both keys are now rejected, naming the
+  spec-level `languageCode` — which is a single setting for the whole build. There is no per-table
+  language and no multi-language labelling; the schema doc now says so, and `references/localization.md`
+  states up front that it covers generative-page code, not Dataverse metadata labels.
+- **The invalid-`languageCode` error names a real LCID.** `"es-ES"` is the natural thing to write and
+  the old message ("must be a positive integer LCID") did not say what to write instead.
+
+[#537]: https://github.com/microsoft/power-platform-skills/issues/537
+
+## [2.6.0]
 
 Business process flows, plus an SDK uptake that changes how business rules fail on an environment
 that cannot host them.
--- a/plugins/model-apps/references/app-spec-schema.md
+++ b/plugins/model-apps/references/app-spec-schema.md
@@ -122,6 +122,19 @@ sample data (incl. multi-parent junction links + status reasons), and publish.
   Must be a positive integer LCID up to 65535 — `1031`, not `"de-DE"` and not `true`. An invalid
   value is rejected by validation, and a caller that bypasses validation gets a warning naming the
   discarded value rather than a silent fall-through.
+  **It is a single, SPEC-LEVEL setting — one language for the whole build.** There is no per-table
+  language and no multi-language labelling:
+  - `languageCode` applies to **every** artifact the build creates. You cannot author one table in
+    Spanish and another in English. The LCID is baked into the SDK at construction time, so a second
+    language would need a second build.
+  - A Dataverse label carries **one** translation here, not several. You cannot give a table both
+    `Customer` and `Cliente` — the SDK's label serializer emits a single `LocalizedLabel`.
+
+  Writing `languageCode` or `localizedLabels` on an **entity** or a **column** is therefore rejected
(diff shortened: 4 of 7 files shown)


akshaymaloo:
## Closing as a duplicate of #539 — #539 is the better fix, and mine has a correctness gap

I wrote this PR, so I tried to compare on evidence rather than preference. #539 wins on a
measurement, not on taste.

### The deciding fact: this PR leaves the bug fully live

`entities[]` is accepted by **two** entry points. I only guarded one. Measured on this branch:

```
entities[].languageCode  = 3082          via provision-input  ->  ok=true  (silently dropped)
entities[].localizedLabels = {...}       via provision-input  ->  ok=true  (silently dropped)
```

So with this PR merged, the exact scenario #537 was filed for is **still reproducible** through the
genpage provisioning path. #539 closes both, sharing one `ENTITY_KEYS` definition between them —
which is also the right call: two entry points disagreeing about what a table key *is* would be
worse than either rule alone.

This PR also introduces a **message inconsistency** I missed: it improves the invalid-LCID text in
`app-spec.js` but leaves `provision-input.js:38` on the old terse string.

### #539 is a superset on every other axis too

- **Catches the whole class, not two keys.** A full 15-key allow-list also rejects `pluralname` and
  every other typo that silently drops today. Mine only ever caught the two keys I happened to think of.
- **Crash-safety** — a revoked `Proxy` and a throwing `ownKeys` trap now produce structured errors
  instead of escaping the validator, preserving the "return problems, never throw" contract.
- **More coverage** — 2046 vs my 2039, plus `AGENTS.md`.

### I verified #539's main risk independently

An allow-list that is too narrow breaks valid specs — a worse failure than the bug. I checked this
myself rather than trusting the claim: validated **every spec-shaped JSON in the repo** (7 files,
both `plan` and `deploy` profiles) against #539's validator. **Zero** allow-list rejections. The
union of entity keys actually used across repo specs is 9 — `columns, description, displayName,
iconDescription, pluralName, primaryAttribute, quickCreate, schemaName, vectorIcon` — a strict
subset of the 15 allowed. Full suite passes on that branch.

I also agree with #539's call **not** to alias BCP-47 tags to LCIDs: `es-ES` is 3082 or 1034
depending on sort order, and a wrong guess would not fail — it would label everything in the wrong
language, which is the same silent-corruption class the fix exists to remove.

Nothing here is worth salvaging into #539. Closing.

PR #562, opened 2026-09-11
model-apps 2.7.1: authored page keys, in-tag comments, value-less view operators, a lint CLI, and the session-start hook warning

### LINKED ISSUE #542: [Bug] promote-intent-pages rejects a valid page: a `//` comment inside a JSX opening tag desyncs source-literals.js
### Plugin

model-apps

### Plugin Version

2.5.1

### Skill / Command

app-builder (generate-pages / `scripts/promote-intent-pages.js`)

### Bug Description

`scripts/lib/source-literals.js` does not enter comment mode for a `//` line comment that appears **inside a JSX opening tag**. An apostrophe in such a comment is therefore treated as the start of a string literal, which swallows the code that follows.

The practical effect is that `promote-intent-pages.js` rejects a **valid, complete** generated page with two misleading findings:

```
no real `export default` (or `export { X as default }`) — the host cannot mount the page
unbalanced brackets — the file looks truncated
```

The file is neither truncated nor missing its default export. Because promotion is transactional, one such page blocks the whole batch and the build cannot proceed.

This is the false-negative direction the library's own header comment names as the worse of the two:

> accepting text as code -> prose is promoted as a page, permanently marked implemented
> rejecting code as text -> a valid generated page is refused and the user is blocked mid-build
>
> The second is worse, which is why this tracks JSX properly instead of running regexes over the raw file.

A `//` comment between attributes in a JSX opening element is valid TSX and is accepted by Babel/TypeScript. It is also a natural thing for a generator to emit when explaining an attribute.

### Steps to Reproduce

Run this against the bundled library — no Dataverse environment needed:

```js
const { hasUnbalancedBrackets, hasDefaultExport } = require('./scripts/lib/source-literals.js');

const base = (comment) =>
  'const A = () => {\n' +
  '  return (\n' +
  '    <button\n' +
  '      type="button"\n' +
  (comment ? '      ' + comment + '\n' : '') +
  '      className={x}\n' +
  '    >hi</button>\n' +
  '  );\n' +
  '};\n' +
  'export default A;\n';

for (const [label, c] of [
  ['no comment',            null],
  ['in-tag // no apostrophe', "// wins over the base border shorthand"],
  ['in-tag // WITH apostrophe', "// wins over Griffel's atomic output"],
]) {
  const s = base(c);
  console.log(label.padEnd(28), 'unbalanced=', hasUnbalancedBrackets(s), 'defaultExport=', hasDefaultExport(s));
}
```

Output:

```
no comment                   unbalanced= false defaultExport= true
in-tag // no apostrophe      unbalanced= false defaultExport= true
in-tag // WITH apostrophe    unbalanced= true  defaultExport= false   <-- valid TSX, rejected
```

Only the apostrophe-bearing in-tag comment fails. The same comment in ordinary **code** position parses correctly, which is why this is easy to miss.

### Expected Behavior

A `//` line comment inside a JSX opening tag is blanked as a comment, exactly as it is in code position. `hasUnbalancedBrackets` returns `false` and `hasDefaultExport` returns `true` for all three cases above.

### Actual Behavior

Inside a JSX opening tag the `//` is not recognised as starting a comment, so the apostrophe opens a phantom string literal. Brackets after that point are miscounted and the real `export default` is blanked, so `promote-intent-pages.js` exits non-zero and reports the page as truncated.

Real failure from a six-page build (the page was 1467 lines with `export default GeneratedComponent;` present on the last line):

```
{
  "ok": false,
  "promoted": [],
  "failed": [
    {
      "index": 0,
      "key": "my-work",
      "file": "my-work.tsx",
      "problems": [
        "no real `export default` (or `export { X as default }`) — the host cannot mount the page",
        "unbalanced brackets — the file looks truncated"
      ]
    }
  ]
}
promotion aborted — 1 of 6 page(s) failed validation; app-spec.json was NOT modified
```

Workaround used: reword the comment to remove the apostrophe. No code change was needed — only comment prose.

### Environment

- OS: Ubuntu 24.04.4 LTS (Linux)
- Claude Code: 2.1.265
- PAC CLI: 2.11.2
- Node: v22.23.2

---
🤖 *This issue was created using the `/report-issue` skill.*

### LINKED ISSUE #545: [Bug] spec-lint passes specs the build rejects: missing top-level `schemaVersion`, and Choice labels in `in`/`not-in` filters
### Plugin

model-apps

### Plugin Version

2.5.1

### Skill / Command

app-builder (`scripts/lib/spec-lint.js`, `scripts/lib/app-spec.js`)

### Bug Description

`spec-lint.js` reports `ok: true, errors: 0` on specs that `build-model-app.js` then rejects. The skill documentation presents the lint as the "guardrail lint (hard gate)", so a clean lint reads as "this spec is buildable" — but two classes of defect are invisible to it. Both were hit on one real app.

**Case 1 — a missing top-level `schemaVersion` silently breaks every page cross-reference.**

`validateAppSpec` derives v2-ness from the **top-level** field (`scripts/lib/app-spec.js`):

```js
const isV2 = (spec.schemaVersion || 0) >= 2;
```

A spec whose `pages[]` entries each carry `"schemaVersion": 2` but which has **no top-level** `schemaVersion` is therefore treated as legacy: `pageKeysSet` is never populated, and page references fall back to matching by **name**. Every `navigatesTo[].targetKey` and every `appShell` `page` subarea then fails to resolve, producing a wall of errors that all point away from the actual cause:

```
page 'x': navigatesTo target 'y' is not a known page key
sitemap subArea references unknown page 'y' (declare it in pages[])
page 'x' is not placed in the sitemap — every page must be an appShell subarea
```

Nothing mentions `schemaVersion`. `spec-lint.js` passes the same spec with 0 errors.

**Case 2 — Choice **labels** are accepted in an `in` / `not-in` view filter, then rejected by Dataverse.**

Label-to-value resolution works for `op: "eq"` (`"value": "Organisation"` builds fine) but **not** for the `in` / `not-in` `values` array. The lint accepts both, and the build fails at the `views` phase with a type error from the platform:

```
✗ view "Accounts Under Review" — HTTP 400: The column, fetchxml, has invalid fetch. Error: An
exception System.FormatException was thrown while trying to convert input value '...' to attribute
'account.xx_datastatus'. Expected type of attribute value: System.Int32.
```

The inconsistency between `eq` and `in` is the trap: the same label works in one operator and not the other, so it looks like a data problem rather than an unimplemented code path.

### Steps to Reproduce

**Case 1**

1. Author a spec with `pages[]` where each page has `"schemaVersion": 2` and a `key`, referenced from `appShell` subareas and from `navigatesTo[].targetKey`. Omit the **top-level** `schemaVersion`.
2. `node -e "const{lintAppSpec}=require('./scripts/lib/spec-lint.js');const r=lintAppSpec(require('./app-spec.json'));console.log(r.ok, (r.errors||[]).length)"` -> `true 0`
3. `node scripts/build-model-app.js --env <envUrl> --spec @app-spec.json` (dry run) -> fails with one "not a known page key" / "unknown page" / "not placed in the sitemap" error per page.
4. Add `"schemaVersion": 2` at the top level. The same dry run now succeeds unchanged.

**Case 2**

1. Add a view filter on a Choice column using labels with `in`:
   ```jsonc
   { "attr": "xx_datastatus", "op": "in", "values": ["Unverified", "Under Review"] }
   ```
2. Lint -> `ok: true, errors: 0`.
3. `--apply` -> the `views` phase halts with the `System.FormatException` above.
4. Replacing the labels with their numeric option values builds fine.

### Expected Behavior

- **Case 1:** either `spec-lint.js` errors when `pages[]` use v2 features (`key`, `navigatesTo`, per-page `schemaVersion: 2`) while the top-level `schemaVersion` is missing, or `validateAppSpec` infers v2 from the pages themselves. Any message naming `schemaVersion` would have made this a one-minute fix instead of a source dive.
- **Case 2:** resolve Choice labels for `in` / `not-in` exactly as for `eq` — or, if that is deliberately unsupported, have the lint reject a non-numeric entry in `values` on a Choice column and say so.

### Actual Behavior

Both specs lint clean and fail at build time with errors that do not name the real cause.

Case 1, full dry-run output for a six-page spec (13 errors, none naming `schemaVersion`):

```
{"ok":false,"errors":[
 "page 'a': navigatesTo target 'b' is not a known page key",
 "page 'c': navigatesTo target 'b' is not a known page key",
 "page 'b': pageInput declares '...' but no page navigates to it with that data",
 "sitemap subArea references unknown page 'd' (declare it in pages[])",
 ... one per page ...
 "page 'a' is not placed in the sitemap — every page must be an appShell subarea"
]}
```

Case 2 affected 5 views in one spec. Note the numeric option values are assigned per publisher, so a spec that hardcodes them is no longer portable between environments — which is an argument for resolving labels in `in` rather than requiring numbers.

### Environment

- OS: Ubuntu 24.04.4 LTS (Linux)
- Claude Code: 2.1.265
- PAC CLI: 2.11.2
- Node: v22.23.2

---
🤖 *This issue was created using the `/report-issue` skill.*

### LINKED ISSUE #546: [Bug] `eq-businessid` / `ne-businessid` missing from spec-lint NO_VALUE_OPS
### Plugin

model-apps

### Plugin Version

2.5.1

### Skill / Command

app-builder (`scripts/lib/spec-lint.js`)

### Bug Description

`NO_VALUE_OPS` in `scripts/lib/spec-lint.js` is missing the `eq-businessid` and `ne-businessid` FetchXML operators, so a view filter using either is rejected as needing a `value` it must not have.

These are valid value-less FetchXML condition operators — the business-unit equivalents of `eq-userid` / `ne-userid`, which **are** in the set. They are the natural way to express "rows owned by my business unit", which is a common view for a multi-business-unit app.

Current set (`scripts/lib/spec-lint.js`, around line 12):

```js
const NO_VALUE_OPS = new Set(['null', 'not-null', 'eq-userid', 'ne-userid', 'eq-useroruserteams', 'eq-userteams',
  'today', 'yesterday', 'tomorrow', 'this-week', 'last-week', 'next-week', 'this-month', 'last-month', 'next-month',
  'this-year', 'last-year', 'next-year', 'this-fiscal-year', 'last-seven-days', 'next-seven-days']);
```

`eq-businessid` and `ne-businessid` are absent.

### Steps to Reproduce

Author a view filter using the operator and run the lint:

```jsonc
{
  "name": "My Business Unit's Records",
  "entity": "account",
  "filters": [ { "attr": "owningbusinessunit", "op": "eq-businessid" } ]
}
```

```bash
node -e "const{lintAppSpec}=require('./scripts/lib/spec-lint.js');console.log(JSON.stringify(lintAppSpec(require('./app-spec.json')),null,2))"
```

### Expected Behavior

The filter lints clean — `eq-businessid` and `ne-businessid` take no `<value>`, so omitting one is correct.

### Actual Behavior

The lint reports an error for a missing value, blocking a valid spec.

None beyond the lint error itself. Workaround used: express the boundary with `eq-useroruserteams` on `ownerid` instead.

### Suggested Fix

Add `'eq-businessid', 'ne-businessid'` to `NO_VALUE_OPS`. It may be worth auditing the set against the full FetchXML value-less operator list at the same time — e.g. `eq-userlanguage`, `olderthan-x-*` and the remaining fiscal-period operators — since each omission fails the same way.

### Environment

- OS: Ubuntu 24.04.4 LTS (Linux)
- Claude Code: 2.1.265
- PAC CLI: 2.11.2
- Node: v22.23.2

---
🤖 *This issue was created using the `/report-issue` skill.*

### LINKED ISSUE #555: `hooks.json` `_comment` key emits a warning on every Claude Code session start
### Plugin

model-apps

### Plugin Version

2.5.1

### Skill / Command

_No response_

## What happens

Claude Code prints this on startup, on every session:

```
model-apps: hooks.json: unknown key "_comment" ignored
```

`plugins/model-apps/hooks/hooks.json` starts with a `_comment` key holding a ~900-character description of what the hooks do. Claude Code validates `hooks.json` against a closed schema, so the key is rejected and warned about rather than silently accepted. The hooks themselves work fine — this is noise only, but it appears in the terminal at every session start.

`plugins/mobile-apps/hooks/hooks.json` carries the same key and presumably produces the same warning for anyone who installs that plugin. Those were the only two of the eight plugins with it at the commit below.

## Reproduce

1. Add the `power-platform-skills` marketplace and install `model-apps`.
2. Start Claude Code.
3. The warning appears in the startup output.

### Expected Behavior

No warning

### Actual Behavior

Warning

```shell
## Suggested fix

Move the description out of the JSON — a `hooks/README.md` beside `hooks.json`, or the plugin's `CHANGELOG.md` / main `README.md`. JSON has no comment syntax and the schema is closed, so there is no in-file place for it.
```

## Environment

- Marketplace commit: `9de213bc33abc39d029f4970c153aa32a42af5ad` (2026-09-01)
- `model-apps` version: 2.5.1
- Claude Code on Windows 11

### LINKED ISSUE #558: [Bug] hooks.json: unknown key "_comment" triggers a startup warning in Claude Code
### Plugin

model-apps (also affects mobile-app)

### Plugin Version

model-apps 2.6.1, mobile-app 0.3.1

### Skill / Command

N/A — plugin hook loading at session start

### Bug Description

Both plugins ship a `hooks/hooks.json` whose top-level object contains a documentation key `"_comment"`. Claude Code validates the hooks file schema and reports unknown keys, so every session start shows:

> ● model-apps: hooks.json: unknown key "_comment" ignored

The warning is harmless (the key is ignored), but it appears on every session start and suggests a plugin misconfiguration to the user.

### Steps to Reproduce

1. Install the `model-apps` (or `mobile-app`) plugin from the `power-platform-skills` marketplace.
2. Start a new Claude Code session.
3. Observe the warning printed during startup.

### Expected Behavior

No warning at session start; hooks.json contains only schema-conformant keys.

### Actual Behavior

`● model-apps: hooks.json: unknown key "_comment" ignored` is shown at every session start.

Affected files/keys:
- `model-apps/hooks/hooks.json` — top-level `"_comment"` (long description of the hook design)
- `mobile-app/hooks/hooks.json` — top-level `"_comment"` ("Telemetry-only start hooks. ...")

Suggested fix: move the explanatory text into a README or a code comment in the referenced hook scripts, and remove the `"_comment"` key from both hooks.json files.

### Environment

- OS: Windows 11 Enterprise (10.0.26200)
- Claude Code: current release as of 2026-09-11

---
🤖 *This issue was created using the `/report-issue` skill.*

### LINKED ISSUE #560: [Enhancement] no CLI entry point for lintAppSpec / validateAppSpec (lint-only mode)
**Plugin:** model-apps
**Plugin Version:** 2.6.1
**Skill / Command:** app-builder (`scripts/lib/spec-lint.js`, `scripts/lib/app-spec.js`)

**Bug Description:**
There is no standalone way to lint or validate an app-spec from the command line. `spec-lint.js`'s `lintAppSpec` and `app-spec.js`'s `validateAppSpec` are library exports with no `require.main` entry point, and `build-model-app.js --help` names no lint-only flag. A headless author (a CI job, or an agent authoring the spec outside the interactive skill flow) has to call the exported functions via `node -e`.

**Steps to Reproduce:**
1. Author an app-spec by hand.
2. Look for a lint entry point: `node scripts/build-model-app.js --help`; `ls scripts/` for a `lint-*.js`.
3. Fall back to `node -e "const {lintAppSpec}=require('./scripts/lib/spec-lint.js'); ..."`.

**Expected Behavior:**
A `--lint-only` flag on `build-model-app.js`, or a `scripts/lint-app-spec.js`, that runs `validateAppSpec` and `lintAppSpec` and exits non-zero on errors.

**Actual Behavior:**
No CLI surface; only the interactive skill flow calls the functions.

**Relevant Logs / Screenshots:** None.

**Environment:** Windows 11, PAC CLI current, Claude Code current. Low priority — a `node -e` workaround exists.

---
🤖 This issue was created using the model-apps plugin's report-issue format.

akshaymaloo:
Closes four open issues an author hits **before ever reaching an environment**, plus the startup warning every user of two plugins sees.

Fixes #542 · Fixes #545 · Fixes #546 · Fixes #555 · Fixes #558 · Fixes #560

### #545 — hand-authored page keys were silently overwritten
`migrateAppSpec` mints a stable `key` per page by slugifying the page `name`, and it ran whenever the **top-level** `schemaVersion` was < 2. A real spec declared `schemaVersion: 2` on each **page**, carried hand-authored keys, and referenced pages **by key** (`navigatesTo[].targetKey`, `appShell` subareas) — but omitted the top-level field. Migration therefore ran and overwrote every authored key with a name-slug. Pass 2 only rewrites refs that match a page *name*, so every key-based ref was left dangling and validation emitted one `not a known page key` / `unknown page` error per page — **none of which named the cause**. Adding the top-level field "fixed" it only by making migration a no-op.

Reproduced before fixing; the authored keys `orders` / `order-card` came back as `all-open-orders` / `order-details-view` while the refs stayed put.

**Fix:** a new pass 0 reserves every authored key, and pass 1 mints only for keyless pages (so a minted slug cannot steal a key a later page authored). A malformed authored key is preserved deliberately, so `validateAppSpec` names it instead of this pass silently replacing it.

### #546 — `eq-businessid` / `ne-businessid` rejected
Both are value-less FetchXML operators — the business-unit equivalents of `eq-userid` / `ne-userid` — so the lint demanded a `value` they must not carry. `NO_VALUE_OPS` is now the documented value-less set (also adding `eq-userlanguage`, the user-hierarchy operators, and the relative fiscal-period ones), checked against the Learn operator reference.

### #555 / #558 — `unknown key "_comment" ignored` at every session start
`model-apps` and `mobile-apps` each shipped a `_comment` documentation key in `hooks/hooks.json`. The host validates that file against a **closed** schema, so it printed a warning on **every launch** — harmless, but it reads as a plugin misconfiguration. The prose moved to `hooks/README.md`, and a new repo-wide `scripts/validate-hooks-manifests.js` fails any manifest with an unrecognised top-level key (allow-list: `hooks`, `description`).

This guard matters more than the one-line deletion: the symptom fails no test and no build, so nothing else in CI or review can catch it. JSON has no comment syntax, and `_comment` is exactly what the next contributor will reach for.

### #542 — a comment inside a JSX opening tag made a valid page look truncated
`source-literals.js` tracks JSX properly rather than running regexes over the raw file, but `jsxTag` mode had **no comment handling at all**. A `//` or `/* */` comment between attributes is valid TSX and a natural thing for a generator to emit when explaining an attribute; an apostrophe in the prose (`// wins over Griffel's atomic output`) was therefore read as an attribute-value quote. That scanner deliberately does not stop at a newline — a JSX attribute value may legitimately span lines — so it ran to **end of file**, blanking the real `export default` and miscounting every bracket after it. `promote-intent-pages` then rejected a complete page as truncated, and because promotion is **transactional**, one such page blocked the whole batch.

This is the false-positive direction the file's own header calls the worse of the two.

Reproduced first. The fix also covers **block** comments, which fail identically and which the report did not mention. I probed the opposite direction too — a URL inside an attribute value is still consumed as a string (the comment branches never see it), JSX text is untouched, a multi-line attribute value is not truncated, and a degenerate unterminated block comment does not crash.
### #560 — no CLI for the two gates
`validateAppSpec` and `lintAppSpec` were library exports with no entry point, so a headless author or CI job had to call them through `node -e` — which is what the docs recommended, and which linted the file **as written** rather than the migrated shape (the very trap in #545). New `scripts/lint-app-spec.js` runs migrate → validate → lint, tags findings `schema:` / `lint:`, and exits non-zero on errors. `--strict` also fails on warnings; `--json` emits the report.

## Review changed this PR

An adversarial review caught a **High-severity defect I introduced**: `lint-app-spec.js` hardcoded the **deploy** validation profile, but the documented gates run it while pages are still `intent` and jobs still lack privileges. Measured — a normal work-in-progress spec failed with `page 'ov': must be implemented ... for a deploy build`, exit 1. My doc change would have broken the authoring flow at both gates. `--profile` now defaults to `plan` (what the builder's own dry run uses); `--profile deploy` gates a final spec.

Three further findings, all verified and fixed:
- Once authored keys survive, a key that equals a **different page's name** was silently retargeted by the pass-2 rewrite — and still validated, because the substitute is itself a valid key. An exact key match now wins.
- `validateAppSpec`'s own warnings were dropped, making the CLI quieter than the build it stands in for.
- `--strict --json` emitted `ok: true` next to exit 1.

One reported category was checked and confirmed clean (no added operator takes a value).

## Verification

| | |
|---|---|
| model-apps suite | **2277 / 2277** |
| mobile-apps suite | **246 pass, 1 pre-existing skip, 0 fail** |
| model-apps evals | **159 / 159** |
| repo validator tests | **71 / 71** |
| repo validators (8) | **all exit 0** |

Both new guards were **mutation-tested**, not just asserted:
- disabling the key-reservation pass fails exactly the one test that covers it; restoring it passes
- disabling the key-precedence rewrite fails exactly its test
- re-adding `_comment` makes the new validator exit 1 with a precise message; removing it exits 0

Not live-tested against a Dataverse environment — every change is in pure spec-processing, linting, docs, or a CI guard, and none of it issues a platform call.

`model-apps` 2.7.1 · `mobile-apps` 0.3.2.

### DIFF
--- a/.github/workflows/validate-repository-metadata.yml
+++ b/.github/workflows/validate-repository-metadata.yml
@@ -32,6 +32,15 @@ jobs:
             - name: validate-telemetry-ikeys
               run: node scripts/validate-telemetry-ikeys.js
 
+            # A hooks.json key the host does not recognise fails nothing — it just
+            # prints "unknown key ... ignored" in every user's terminal at every
+            # session start, so no test and no reviewer catches it.
+            - name: test-hooks-manifests-validator
+              run: node --test scripts/tests/validate-hooks-manifests.test.js
+
+            - name: validate-hooks-manifests
+              run: node scripts/validate-hooks-manifests.js
+
             - name: test-secure-process-execution-validator
               run: node --test scripts/tests/validate-secure-process-execution.test.js
 
--- a/plugins/mobile-apps/.claude-plugin/plugin.json
+++ b/plugins/mobile-apps/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
     "name": "mobile-app",
-    "version": "0.3.1",
+    "version": "0.3.2",
     "description": "Build and deploy Power Apps code apps for mobile using Expo (React Native), with native device functionality and Power Platform connectors",
     "author": {
         "name": "Microsoft",
--- a/plugins/mobile-apps/.plugin/plugin.json
+++ b/plugins/mobile-apps/.plugin/plugin.json
@@ -1,6 +1,6 @@
 {
     "name": "mobile-app",
-    "version": "0.3.1",
+    "version": "0.3.2",
     "description": "Build and deploy Power Apps code apps for mobile using Expo (React Native), with native device functionality and Power Platform connectors",
     "author": {
         "name": "Microsoft",
--- a/plugins/mobile-apps/hooks/README.md
+++ b/plugins/mobile-apps/hooks/README.md
@@ -1,0 +1,15 @@
+# mobile-apps hooks
+
+What `hooks.json` wires up, and why.
+
+Telemetry-only start hooks. They never validate, mutate, or block tool calls. File validation stays
+owned by each workflow through `scripts/validate-mobile-files.js`.
+
+This prose lived in a `_comment` key inside `hooks.json` itself. Claude Code validates that file
+against a closed schema and prints `mobile-apps: hooks.json: unknown key "_comment" ignored` at
+**every session start**, which reads as a plugin misconfiguration to the user (#555, #558). The
+documentation belongs here instead; `scripts/validate-hooks-manifests.js` keeps it from drifting
+back.
+
+Telemetry is fail-closed and honours the `POWER_PLATFORM_SKILLS_TELEMETRY_MOBILE_APP_OPTOUT`
+environment variable, which disables transmission for automation and CI.
--- a/plugins/mobile-apps/hooks/hooks.json
+++ b/plugins/mobile-apps/hooks/hooks.json
@@ -1,5 +1,4 @@
 {
-  "_comment": "Telemetry-only start hooks. They never validate, mutate, or block tool calls. File validation stays owned by each workflow through scripts/validate-mobile-files.js.",
   "hooks": {
     "PreToolUse": [
       {
--- a/plugins/model-apps/.claude-plugin/plugin.json
+++ b/plugins/model-apps/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "model-apps",
-  "version": "2.7.0",
+  "version": "2.7.1",
   "description": "Build model-driven Power Apps end to end: /app-builder authors whole apps (tables, relationships, forms, views, charts, security roles, app + sitemap) from a natural-language intent, and /genpage builds generative pages with specialist agents for planning, entity creation, and parallel code generation. Requires PAC CLI > 2.10.0 and Azure CLI (`az`). See CHANGELOG.md for v1.x -> v2.x migration.",
   "author": {
     "name": "Microsoft",
--- a/plugins/model-apps/.plugin/plugin.json
+++ b/plugins/model-apps/.plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "model-apps",
-  "version": "2.7.0",
+  "version": "2.7.1",
   "description": "Build model-driven Power Apps end to end: /app-builder authors whole apps (tables, relationships, forms, views, charts, security roles, app + sitemap) from a natural-language intent, and /genpage builds generative pages with specialist agents for planning, entity creation, and parallel code generation. Requires PAC CLI > 2.10.0 and Azure CLI (`az`). See CHANGELOG.md for v1.x -> v2.x migration.",
   "author": {
     "name": "Microsoft",
--- a/plugins/model-apps/AGENTS.md
+++ b/plugins/model-apps/AGENTS.md
@@ -99,6 +99,14 @@ the pipeline and delegates each script's **behavioral spec** to the entries belo
 - **`scripts/lib/spec-lint.js`** — pure App Spec guardrail (`lintAppSpec → { ok, errors,
   warnings }`): errors block the plan gate (e.g. the relationship-name-vs-lookup-name
   collision Dataverse rejects), warnings teach.
+- **`scripts/lint-app-spec.js`** — the CLI surface for both gates, for a headless author or a CI
+  job (#560). Runs the same three steps every deploying CLI runs on load — `migrateAppSpec` →
+  `validateAppSpec` → `lintAppSpec` — and exits non-zero on errors (`--strict` also fails on
+  warnings). Prefer it over calling the library through `node -e`: linting the file **as written**
+  rather than the migrated shape is what let a spec pass the lint and then fail the build (#545).
+  `--profile` defaults to **`plan`**, not `deploy`, because the authoring flow runs this while pages
+  are still intents and jobs still lack privileges — a deploy default rejects a normal
+  work-in-progress spec. Gate a final, deployable spec with `--profile deploy`.
 - **`scripts/build-model-app.js` → `scripts/lib/sdk-build.js`** — the deterministic, **idempotent**
   build engine, run after approval. Runs in two engine invocations (staged flow): (1) `--stage data
   --apply` materializes tables + columns + relationships (solution·data-model only; sample-data
--- a/plugins/model-apps/AGENTS.md
+++ b/plugins/model-apps/AGENTS.md
@@ -100,13 +100,13 @@ the pipeline and delegates each script's **behavioral spec** to the entries belo
   warnings }`): errors block the plan gate (e.g. the relationship-name-vs-lookup-name
   collision Dataverse rejects), warnings teach.
 - **`scripts/lint-app-spec.js`** — the CLI surface for both gates, for a headless author or a CI
-  job (#560). Runs the same three steps every deploying CLI runs on load — `migrateAppSpec` →
-  `validateAppSpec` → `lintAppSpec` — and exits non-zero on errors (`--strict` also fails on
+  job (#560). Runs `migrateAppSpec` → `validateAppSpec` (what the build runs on load) → `lintAppSpec`
+  (guardrails the builder does **not** run), and exits non-zero on errors (`--strict` also fails on
   warnings). Prefer it over calling the library through `node -e`: linting the file **as written**
   rather than the migrated shape is what let a spec pass the lint and then fail the build (#545).
   `--profile` defaults to **`plan`**, not `deploy`, because the authoring flow runs this while pages
-  are still intents and jobs still lack privileges — a deploy default rejects a normal
-  work-in-progress spec. Gate a final, deployable spec with `--profile deploy`.
+  are still intents — a deploy default rejects a normal work-in-progress spec. `plan` relaxes only
+  that rule. Gate a final, deployable spec with `--profile deploy`.
 - **`scripts/build-model-app.js` → `scripts/lib/sdk-build.js`** — the deterministic, **idempotent**
   build engine, run after approval. Runs in two engine invocations (staged flow): (1) `--stage data
   --apply` materializes tables + columns + relationships (solution·data-model only; sample-data
--- a/plugins/model-apps/CHANGELOG.md
+++ b/plugins/model-apps/CHANGELOG.md
@@ -5,7 +5,47 @@ All notable changes to the **model-apps** plugin.
 Entries are deliberately short: what changed and why it matters to you. The reasoning,
 evidence and trade-offs behind a change live in its PR, in `docs/`, or in the linked issue.
 
-## [Unreleased] — 2.7.0
+## [Unreleased] — 2.7.1
+
+Three defects an author hits before reaching an environment, and a session-start warning.
+
+### Added
+
+- **`scripts/lint-app-spec.js` — lint and validate a spec without touching an environment** ([#560]).
+  `validateAppSpec` (the gate `--apply` enforces) and `lintAppSpec` (the authoring guardrails) were
+  library exports with no entry point, so a headless author or a CI job had to reach in with
+  `node -e`. The CLI runs the same three steps every deploying CLI runs on load — migrate →
+  validate → lint — tags each finding `schema:` or `lint:`, and exits non-zero on errors.
+  `--profile` defaults to `plan` (pages may still be intents, jobs may still lack privileges), so
+  gate a final, deployable spec with `--profile deploy`. `--strict` also fails on warnings;
+  `--json` emits the report.
+
+### Fixed
+
+- **A spec that declares `schemaVersion` per page but not at the top level no longer breaks every
+  page reference** ([#545]). Migration mints a stable key per page, and it did so *unconditionally* —
+  overwriting hand-authored keys with a slug of the page name. The references hold keys, not names,
+  so nothing rewrote them and validation reported one "not a known page key" / "unknown page" error
+  per page, none of which named the cause. Authored keys now survive migration; a key is minted only
+  for a page that has none, and every authored key is reserved first so a minted one cannot steal it.
+- **`eq-businessid` / `ne-businessid` are accepted in a view filter** ([#546]). Both are value-less
+  FetchXML operators — the business-unit equivalents of `eq-userid` / `ne-userid` — and the lint
+  demanded a value they must not carry. The operator list is now the documented value-less set,
+  which also adds `eq-userlanguage`, the user-hierarchy operators, and the relative fiscal-period
+  ones.
+- **No more `hooks.json: unknown key "_comment" ignored` at every session start** ([#555], [#558],
+  affects `model-apps` and `mobile-apps`). The hooks manifest is validated against a closed schema,
+  so the documentation key it carried was reported as a misconfiguration on every launch. The prose
+  moved to `hooks/README.md`, and a repo-wide CI check now fails any manifest with an unrecognised
+  top-level key — the symptom is otherwise invisible to both tests and review.
+
+[#545]: https://github.com/microsoft/power-platform-skills/issues/545
+[#546]: https://github.com/microsoft/power-platform-skills/issues/546
+[#555]: https://github.com/microsoft/power-platform-skills/issues/555
+[#558]: https://github.com/microsoft/power-platform-skills/issues/558
+[#560]: https://github.com/microsoft/power-platform-skills/issues/560
+
+## [2.7.0]
 
 Multi-language Dataverse labels, granting access on roles this spec does not own, and the
 app-builder defects found while rebuilding a real app into a second environment.
--- a/plugins/model-apps/CHANGELOG.md
+++ b/plugins/model-apps/CHANGELOG.md
@@ -28,6 +28,13 @@ Three defects an author hits before reaching an environment, and a session-start
   so nothing rewrote them and validation reported one "not a known page key" / "unknown page" error
   per page, none of which named the cause. Authored keys now survive migration; a key is minted only
   for a page that has none, and every authored key is reserved first so a minted one cannot steal it.
+- **A `//` or `/* */` comment inside a JSX opening tag no longer makes a valid page look truncated**
+  ([#542]). `jsxTag` mode had no comment handling, so an apostrophe in the comment prose was read as
+  an attribute-value quote — and that scanner does not stop at a newline, because a JSX attribute
+  value legitimately may span lines. It ran to end of file, blanking the real `export default` and
+  miscounting every bracket after it, so `promote-intent-pages` rejected a complete page as
+  truncated. Promotion is transactional, so one such page blocked the whole batch. Block comments
+  had the same defect, which the report did not cover.
 - **`eq-businessid` / `ne-businessid` are accepted in a view filter** ([#546]). Both are value-less
   FetchXML operators — the business-unit equivalents of `eq-userid` / `ne-userid` — and the lint
   demanded a value they must not carry. The operator list is now the documented value-less set,
@@ -39,6 +46,7 @@ Three defects an author hits before reaching an environment, and a session-start
   moved to `hooks/README.md`, and a repo-wide CI check now fails any manifest with an unrecognised
   top-level key — the symptom is otherwise invisible to both tests and review.
 
+[#542]: https://github.com/microsoft/power-platform-skills/issues/542
 [#545]: https://github.com/microsoft/power-platform-skills/issues/545
 [#546]: https://github.com/microsoft/power-platform-skills/issues/546
 [#555]: https://github.com/microsoft/power-platform-skills/issues/555
--- a/plugins/model-apps/CHANGELOG.md
+++ b/plugins/model-apps/CHANGELOG.md
@@ -14,11 +14,10 @@ Three defects an author hits before reaching an environment, and a session-start
 - **`scripts/lint-app-spec.js` — lint and validate a spec without touching an environment** ([#560]).
   `validateAppSpec` (the gate `--apply` enforces) and `lintAppSpec` (the authoring guardrails) were
   library exports with no entry point, so a headless author or a CI job had to reach in with
-  `node -e`. The CLI runs the same three steps every deploying CLI runs on load — migrate →
-  validate → lint — tags each finding `schema:` or `lint:`, and exits non-zero on errors.
-  `--profile` defaults to `plan` (pages may still be intents, jobs may still lack privileges), so
-  gate a final, deployable spec with `--profile deploy`. `--strict` also fails on warnings;
-  `--json` emits the report.
+  `node -e`. The CLI runs migrate → validate → lint, tags each finding `schema:` or `lint:`, and
+  exits non-zero on errors. `--profile` defaults to `plan` (pages may still be intents), so gate a
+  final, deployable spec with `--profile deploy`. `--strict` also fails on warnings; `--json`
+  emits the report.
 
 ### Fixed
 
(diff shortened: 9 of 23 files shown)


akshaymaloo:
### Review response — all 7 findings verified by measurement, all 7 fixed

Thanks — two of these were **real regressions I introduced**, and one disproved a claim I had written into a comment. Every item below was reproduced before it was changed.

| # | Finding | Verdict | Fix |
|---|---|---|---|
| 1 | Key-first rewrite can misresolve a legacy name ref via a **minted** key | **Confirmed** | `fb1bb28d` |
| 2 | Present-but-malformed `key` still silently replaced | **Confirmed** | `fb1bb28d` |
| 3 | Bare `--spec` reports a gate verdict, not a usage error | **Confirmed** | `fb1bb28d` |
| 4 | Bare `--profile` silently falls back to `plan` | **Confirmed** | `fb1bb28d` |
| 5 | `plan` does not actually relax job `privileges[]` | **Confirmed — my comment was wrong** | `fb1bb28d` |
| 6, 7 | "same three steps every deploying CLI runs" overstates parity | **Confirmed** | `fb1bb28d` |

**1 — the sharpest of the set.** I had used "is this string any final page key?" as the key-first test. But a *minted* key is derived from a name, so it collides with a different page's name **by construction**. Measured on the previous commit:

```
pages: [ { name: 'All Orders' }, { name: 'all-orders', navigatesTo: [{ targetKey: 'all-orders' }] } ]
keys       : all-orders, all-orders-2
nav target : all-orders      <- wrong page; the ref names the SECOND page
```

The test is now the set of **authored** keys. A legacy spec has none, so it keeps exactly its previous name-first behaviour — which is the property that was broken.

**2.** `{ key: 42 }`, `{ key: null }`, `{ key: "" }` and `{ key: "   " }` all came out as the minted slug `"orders"`, so `validateAppSpec`'s key error never fired — directly contradicting the preservation the pass documents. A key is now minted only when the page has **no `key` property at all**; a malformed one reaches validation unchanged and is named there, and is never reserved or used to resolve a reference.

**3 / 4.** Both are now usage errors. The `--profile` case is the one with teeth: a CI job that wrote `--profile` without a value would have believed it requested a `deploy` gate and silently got `plan`.

**5.** You were right and my comment was wrong. Measured — `privileges[] is required` fires under **all four** profiles:

```
design      ok = false | ["persona \"Rep\" job \"?\": privileges[] is required", ...]
plan        ok = false | [... same ...]
deploy      ok = false | [... same ...]
structural  ok = false | [... same ...]
```

`plan` relaxes page implementation and nothing else. Corrected in the CLI header, the `lintSpec` doc comment, `AGENTS.md`, `SKILL.md`, `authoring-flow.md` and the CHANGELOG.

**6 / 7.** Confirmed: `build-model-app.js` migrates and validates but never calls `lintAppSpec`. The wording now says so explicitly, and notes that the `lint:` findings are *additional* to what the builder reports — which is the point of running it at the plan gate.

Findings 1 and 2 are covered by new tests in `app-spec-migrate.test.js`; 3 and 4 by a subprocess test in `lint-app-spec.test.js`. **2280/2280** model-apps tests, 159/159 evals, 8/8 repo validators.

akshaymaloo:
Addressed in 53e273e4. Four findings — two confirmed, one **refuted by measurement** and fixed at the severity the evidence supports, one a genuine test gap.

### 1. `--profile=` silently took the default (the open thread) — confirmed, fixed

`parseArgs` preserves `--profile=` as `'`, which passed the "is a string" guard and was then folded into the default by `|| 'plan'`. Only `undefined` may take a default now, in **both** the library (`lintSpec`) and the CLI, so a programmatic caller cannot hit it either.

### 2. Unknown flags silently ignored — confirmed, fixed

`--profle deploy` ran the default `plan` profile and exited 0, reporting success for a gate it never applied. Unknown flags and stray positionals are now usage errors. This CLI ships in this PR, so no caller can depend on the old behaviour.

### 3. Value on a value-less operator — **premise refuted; fixed as a warning, not an error**

The finding asked me to reject `{ op: 'eq-businessid', value: ... }` because `sdk-build.js` forwards the value and "the platform rejects it". I probed the Dataverse Web API directly rather than take that on trust:

| condition | result |
|---|---|
| `eq-businessid value="00000000-0000-0000-0000-000000000000"` | HTTP 200, 1 row |
| `eq-businessid` (no value) | HTTP 200, 1 row |
| `this-year value="1999"` | HTTP 200, 1 row |
| `this-year` (no value) | HTTP 200, 1 row |

The platform neither rejects the condition nor honours the value — it **ignores** it. A zero-GUID business unit returns the same row as no value at all. So:

- **An error would be wrong**: it would block specs that build and run correctly today.
- **Silence would also be wrong**: the filter does not do what its `value` plainly says.

Hence a **warning**, and `sdk-build.js` is deliberately unchanged — there is no platform failure to prevent, and changing emission would be a behaviour change with no defect behind it.

### 4. Regression cases for every new operator — confirmed, fixed

You were right that the existing test covered 6 of 31 operators. Worth naming the sharper version of the point: a test that iterates `NO_VALUE_OPS` to check `NO_VALUE_OPS` **cannot catch a typo inside it** — a misspelled entry simply goes untested. The expected list is now transcribed independently from the documented operator table and compared in **both** directions, so a typo reintroduces the #546 false positive *and* an extra value-taking entry (which would silently disable the missing-value check) both fail. All 31 names were verified verbatim against the operator table; there were no typos.

### 5. `process.exit()` truncating a piped report — not reproducible; changed anyway, not claimed as a fix

I could not reproduce this. A 5 MB write over a pipe survives `process.exit()` intact on Windows, and Node documents pipes as **synchronous on both Windows and POSIX** — so a `--json | jq` CI consumer was never at risk. The documented asynchronous case is a **TTY on Windows**, i.e. an interactive run. I switched to `process.exitCode` because it is the correct idiom for that case, but the comment says what was measured rather than repeating the premise, and I deliberately did **not** add a test — no failing case could be produced, and a test that passes either way is not a guard.

### Verification

All 7 new guards were mutation-tested: disabled one at a time, each caught by exactly the intended test (including the `NO_VALUE_OPS` typo, caught by the new set-equality test). 2287/2287 model-apps tests (+7), 159/159 evals, 8/8 repo validators.

akshaymaloo:
Peer review round 2 (independent adversarial review of 53e273e4) found ten issues. All addressed in c09e352c — including **two regressions I introduced in that commit** and **one claim of my own that the review correctly falsified**.

### My regressions, caught by the review

1. **The new positional guard destroyed the diagnostic for an unquoted Windows path with spaces.** The guard fired *before* `--spec` was inspected, so `--spec C:\Users\x\OneDrive - Contoso\app-spec.json` (the documented flow asks for an absolute path) produced `unexpected argument(s): -, Contoso\app-spec.json — the spec is named with --spec` — which the caller *had* done. The pre-commit message named the truncated file. It now reports what `--spec` actually parsed to, plus a quoting hint.

2. **One warning covered both `value` and `values`, but they are dropped at different points.** `value` is forwarded into the condition and ignored by Dataverse; `values` is only read by the `in`/`not-in` branch, so on a value-less operator it never reaches the platform at all. Saying "Dataverse ignores it" for the `values` case sent the author to the wrong place. Split in two, and the test pins the distinction.

### The claim the review falsified

My previous comment said the expected-operator list was "independently transcribed from the documented table". It was in **the source module's editorial grouping** — a reflow of the very thing it checks. The *data* was correct (all 31 verified against the table), but the claim was not, and a typo present at authoring time would have been copied into both. It is now in the doc table's alphabetical order so the transcription is verifiable by eye.

### New, beyond the original findings

- **`KNOWN_FILTER_OPS`** — the full documented operator table. Without it, a typo'd operator fell through to the missing-value check and was reported as `needs a value`: **actively wrong advice**. The author adds a value, the lint goes green, Dataverse rejects the operator at build time — the #546 shape with an extra step. Now warns with a "did you mean" near match, mirroring the business-rule operator check in `app-spec.js`. Deliberately a **warning**: an incomplete list here would reject a correct spec, which is the exact failure this PR exists to fix.
- **Usage errors emit parseable JSON under `--json`.** The documented invocation is `--spec @<path> --json` and the caller parses stdout — so a usage error left stdout empty and `JSON.parse('')` threw. An agent following the documented flow got a crash instead of the usage message.
- **`--help`** prints usage and exits 0 rather than `unknown flag(s): --help`.
- **The allow-list reads flag names from `argv`**, not `Object.keys(flags)`: `--__proto__` goes through the inherited setter and never becomes an own property, so it would have escaped the allow-list *and* swallowed the next token.
- **Profile rejections are tagged `schema:`** like every other error — `authoring-flow.md` tells the agent to triage on that prefix, and nothing pinned the invariant on the errors side.
- **The operator sets are exported frozen, as arrays** — the live `Set` is shared through the require cache.

### Considered and deliberately not done

Switching the shared `parseArgs` to `Object.create(null)`. It addresses the same `__proto__` hazard, but the exploitable path is already closed above, and the change **broke five existing assertions** in `dataverse-auth.test.js` that compare the flags object shape. Re-pointing a helper used by every CLI in the plugin, for a Low finding already mitigated where it matters, is not a trade worth making in this PR. Noted as a follow-up.

### Verification

All **10** new guards mutation-tested — disabled one at a time, each caught by exactly the intended test (including the `__proto__` hole and the `KNOWN_FILTER_OPS` typo). **2298/2298** model-apps tests, **159/159** evals, **8/8** repo validators. Every documented invocation re-run end to end against a real spec.

akshaymaloo:
All four addressed in 3c89f2ba.

**1. `spec-lint.js` — the warning asserted match semantics that are wrong for much of the set.** Correct, and I should have caught it: the message said the filter "matches the current user/business-unit/period regardless", but `null`/`not-null` test presence and `eq-userlanguage` matches the user's language — neither is a user/BU/period match. The message now states only that the value **has no effect on what the filter matches**, which holds for every operator in the set, and a comment records why the message deliberately stops short of naming semantics.

**2-4. Review attribution in committed test comments.** Agreed, and it is the repo's own rule (`AGENTS.md`: internal review process must not be committed). The rationale is the valuable part and is kept verbatim; the `Review finding (#562)` / `Review finding.` attribution is gone, rewritten as neutral regression contracts. One comment that said a reviewer "correctly called out" the expected-list problem now just states the problem.

Scoped to the files this PR already touches. The same phrasing exists elsewhere in the plugin (`role-privileges.test.js`, and a whole `review-findings.test.js`); I left those alone rather than widening an otherwise-ready diff — worth a separate sweep if you want the convention enforced plugin-wide.

2298/2298 model-apps tests, 8/8 repo validators.

nguyentony95 (approved):

PR #566, opened 2026-09-15
model-apps: download keeps Choice/MultiChoice types (#564); genpage prompts survive multi-line (#565)

### LINKED ISSUE #564: [model-apps][Bug] download drops Choice/MultiChoice column types, so a rebuild into a fresh environment creates them as Text
## What

`download-model-app.js` emits Choice / MultiChoice columns **without a `type`**, because a Choice column's App Spec declaration requires a companion `options[]` or `globalChoice` reference that the download does not read (see the `TYPES_NEEDING_COMPANION_DATA` comment in `scripts/download-model-app.js`).

## Why it matters

Rebuilding the downloaded spec into an environment that **already has the table** is inert. Rebuilding into a **fresh** environment silently creates the column as single-line **Text**, because the data-model phase falls back to `SDK_COLUMN_TYPE[c.type || 'Text']`.

The result is asymmetric and therefore easy to miss: Memo, Money and DateTime round-trip correctly while Choice quietly degrades.

## Live evidence

Downloading a one-table app whose only Choice column was `new_stage`:

```
WARNING: 1 column(s) were captured WITHOUT a type (new_regritem.new_stage) — most likely
Choice/MultiChoice columns, whose options this download does not read. ... rebuilding into a
FRESH environment would create them as single-line Text.
```

The warning is accurate and the behaviour is deliberate, but the underlying gap is fixable.

## The read that closes it (verified live, HTTP 200)

```
EntityDefinitions(LogicalName='<table>')/Attributes/Microsoft.Dynamics.CRM.PicklistAttributeMetadata
  ?$select=LogicalName
  &$expand=OptionSet($select=Name,IsGlobal,Options),GlobalOptionSet($select=Name)
```

Measured against a live environment on `account`: 20 picklist attributes returned. A **local** option set comes back as `OptionSet.IsGlobal === false` with `OptionSet.Options[]`, each option carrying `Value` and `Label.UserLocalizedLabel.Label` — e.g. `accountcategorycode` -> `[{Value:1,Label:"Preferred Customer"},{Value:2,Label:"Standard"}]`. A **global**-bound picklist reports `OptionSet.IsGlobal === true` with the shared set's `Name`, which maps to a `globalChoice` reference rather than inline `options[]`.

`MultiSelectPicklistAttributeMetadata` is the parallel cast for MultiChoice.

## Scope

- Read picklist + multi-select metadata per downloaded entity.
- Local option set -> inline `options[]`; global -> `globalChoice` reference.
- Emit the real `type` once the companion data is present, and drop the affected column from the untyped-column warning.
- Cover both shapes with tests; live-verify a download -> rebuild-into-a-fresh-org round trip.

Filed rather than folded into #563: that PR is already 12 commits and under review, and this is a feature-sized change to the download path rather than a regression from it.

### LINKED ISSUE #565: [model-apps][Enhancement] genpage upload: multi-line prompts are flattened because --prompt-file / --agent-message-file / --compiled-code-file are never used
## What

`pac model genpage upload` accepts three inputs the plugin never supplies:

| pac flag | plugin support |
|---|---|
| `--compiled-code-file` | never passed on any path |
| `--prompt-file` | never passed; the prompt is passed inline via `--prompt` |
| `--agent-message-file` | never passed; passed inline via `--agent-message` |

Confirmed against `pac model genpage upload --help` and by grepping `scripts/lib/genpage-cli.js` plus the `/genpage` skill body.

For clarity, these are NOT gaps: `genpage-cli.js` deliberately omits `--add-to-sitemap` (documented at the top of the file — the SDK owns the sitemap and writes the GenPage subareas), and `--connectors` / `--actions` / `--model` are passed by the `/genpage` skill on its own raw-`pac` deploy path.

## Why `--prompt-file` is the interesting one

`genpage-cli.js` `quoteArg()` currently does this:

```js
const s = String(a).replace(/\r\n|[\r\n]/g, ' ');
```

with the comment: *"Embedded newlines terminate the command line (pac then sees a truncated command and dumps its help), so collapse them to spaces first — downloaded page prompts are multi-line (`Conversation with N prompts:\r\n1. …\r\n2. …`) and would otherwise break the upload on an edit-rebuild."*

So the plugin is **lossily flattening** multi-line prompts to work around a command-line limitation that pac already solves with `--prompt-file`. On an edit-rebuild the stored prompt is a conversation transcript, and today every newline in it becomes a space. Switching to `--prompt-file` would round-trip it intact and delete the workaround.

`--agent-message-file` is the same argument for the agent message.

## `--compiled-code-file`

Optional by design — pac auto-transpiles the TypeScript when it is absent, which is the path the plugin uses and which is live-verified working. The gap only matters for a caller that already has compiled JS and wants to skip transpilation. Lower value than the prompt flags; listed for completeness since it was the last unhandled genpage input carried over from an older internal notes file.

## Suggested scope

1. Pass the prompt and agent message via temp files (`--prompt-file` / `--agent-message-file`) and drop the newline-collapsing in `quoteArg`, with a test proving a multi-line prompt survives an upload -> download round trip.
2. Optionally thread `--compiled-code-file` through the wrapper.

akshaymaloo:
Stacked on #563 (base is that branch, so the diff below is only this work; GitHub retargets to `main` automatically when #563 merges).

Closes #564. Closes #565.

## #564 — download keeps Choice/MultiChoice types

A downloaded Choice column carried **no `type`**, because that declaration obliges a companion `options[]` or `globalChoice` and nothing read one. Rebuilding into the **original** environment reused the real column and hid it; rebuilding into a **fresh** one created single-line **Text**, while Memo, Money and DateTime round-tripped. The asymmetry is what made it dangerous.

Option sets are now read per table through the two picklist metadata **casts**. A **local** set becomes inline `options[]`; a **shared** one becomes a `globalChoice` reference plus a `globalChoices[]` declaration, so a fresh-environment rebuild can create it.

Three things were **measured, not assumed**, and each changed the design:

- **`GlobalOptionSet` is not the global/local discriminator.** It comes back populated for a **local** set too, echoing that set's own name. The obvious reading of the API would have emitted a `globalChoice` reference to a shared set that does not exist, binding the column to nothing. `OptionSet.IsGlobal` is the only honest discriminator, and `OptionSet.Options` is populated for both kinds.
- **A MultiSelectPicklist reports `AttributeType: "Virtual"`.** Only `AttributeTypeName` says `MultiSelectPicklistType`, and the SDK projection does not carry it — so a MultiChoice column looked exactly like the synthetic `<column>name` shadows beside it and was filtered out of `columns[]` **entirely**. The spec lost the **column**, not just its type, and a rebuild never created it. This is strictly worse than what #564 reported, and it was found **only** because the live round trip was run. Membership in the multi-select cast response is now the discriminator; the shadows stay filtered.
- **Inline options keep their localizations; a `globalChoices[]` declaration must be flattened,** because validation rejects a localized shared-set option (Dataverse stores only the base language and reports nothing).

Fail-safe is preserved, which is why the untyped warning survives: a failed or incomplete read still leaves the column untyped — emitting `type: "Choice"` with no companion produces a spec that fails its own validation — but now names the table and the reason. Option **order** is left exactly as Dataverse returned it (display order), because the App Spec assigns `value = 100000000 + index`. The not-round-tripped report no longer lists a global choice the spec now declares.

**Known limit, documented:** the App Spec cannot express option *values*. Labels and order round-trip; a fresh rebuild re-bases the integers to `100000000 + index`.

## #565 — genpage upload passes the prompt by file

The prompt and agent message were passed **inline**, so both hit the newline-collapsing guard that stops a stray newline truncating the Windows command line. A downloaded prompt is a multi-line conversation transcript, so on an **edit-rebuild** every line break silently became a space. They now use `--prompt-file` / `--agent-message-file`.

The collapsing guard is **kept**, contrary to the issue's suggestion: it is no longer lossy for prompts (they never reach it) but is still load-bearing for the remaining inline args — a page `--name` can truncate a command line just as easily. Removing it would turn a lossy-but-working path into a hard failure. The comments were rewritten instead, so they no longer justify themselves with a reason that stopped being true.

Temp files are written once before the retry loop and a `finally` removes them on **every** exit: success, retry-exhaustion throw, and the mid-loop identity-mismatch throw. `upload()` also accepts an optional `compiledCodeFile`; omitted, pac auto-transpiles as before.

## Verification

**Live, against a real environment** (both probe apps and the probe page removed afterwards; teardowns reported 0 failures):

- #564: build → download → rebuild under a **fresh identity** → Dataverse reports `Picklist` and `MultiSelectPicklistType` with options in order. Before the change the same round trip produced Text and a **missing** column. Separately, a stock `contact` went from 0 of 24 picklist columns typed to **24 of 24**, localizations intact, with its one global-bound column emitted as a reference.
- #565: a prompt carrying `\r\n`, a bare `\n`, an embedded quote, a `%` and non-ASCII round-tripped **byte-for-byte** through upload → download, on both the create and update paths. Zero temp directories leaked across live runs, the full suite, and a path that threw after three failed attempts.

**Automated:** full suite **2436/2436**, evals **159/159**, all **9** repo validators green, no real environment identifiers in tracked files.

**Mutation-tested:** every new guard. Two findings worth stating — one mutation initially **survived** because a test bypassed the code path it claimed to cover (the test was rewritten to drive the real read path), and one was a genuinely **equivalent** mutation, now pinned explicitly so the intended precedence cannot be "simplified" away later.


akshaymaloo:
### ⚠️ CI note: the test workflows have not run on this PR

Every workflow in this repo is gated on `pull_request: branches: [main]`. This PR targets `users/akmaloo/dry-run-and-self-parent` (it is stacked on #563 to keep the diff reviewable), so `model-apps-script-tests`, `model-apps-evals` and `validate-keyword-case` were **never triggered**. The `CLEAN` merge state therefore means "nothing ran", **not** "tests passed" — please do not read it as a green suite.

They will run once #563 merges and GitHub retargets this PR to `main`.

In the meantime the CI matrix was reproduced locally, on the axes that exist:

| Check | Result |
|---|---|
| `model-apps` script suite, Node 20 | 2436 / 2436 |
| `model-apps` script suite, Node 22 | 2436 / 2436 |
| `model-apps` evals, Node 20 | 159 / 159 |
| `model-apps` evals, Node 22 | 159 / 159 |
| All 9 repo validators | pass |
| `validate-no-real-environments` | pass (no real identifiers in tracked files) |

Not reproducible locally: the ubuntu and macOS legs of the matrix. That matrix exists because these scripts build filesystem paths and parse CLI output, so path-separator differences are a real regression class — worth a look when the workflows do run. The changes here are metadata reads plus `fs`/`path`/`os` temp-file handling (`path.join`, `os.tmpdir()`), with no hard-coded separators.

akshaymaloo:
Consolidated into #563 rather than shipping as a stack, per review preference.

Both commits (`0eb23325` for #564, `17f8a3fe` for #565) were fast-forwarded onto `users/akmaloo/dry-run-and-self-parent`, so #563 now carries the whole model-apps 2.8.0 release as a single reviewable PR. Nothing is lost — this branch's head and #563's head are the same commit (`17f8a3fe`).

This also closes the CI gap flagged in the comment above: #563 targets `main`, so the path-filtered `model-apps-script-tests` and `model-apps-evals` workflows actually run, which they never did while this PR was stacked on another branch.

PR #568, opened 2026-09-15
fix(model-apps): verify checks app-module TABLE membership, not just the sitemap (AB#6699081)

akshaymaloo:
Stacked on #563 (base is that branch, so the diff below is only this work; GitHub retargets to `main` when #563 merges).

Addresses **AB#6699081**, which was filed as a suspected regression of AB#6612527. **It is not a regression** — and the part that *is* real is the verifier, not the build.

## What I could not reproduce, and the evidence

The report says the app-shell update writes the sitemap but omits the new table from the app module. I could not reproduce that through any supported path. Measured live:

| Scenario | Result |
|---|---|
| **Create** path (the 6612527 fix) | Intact — 7/7 real-bundle contract tests pass; a fresh 2-table app returns two **distinct** type-1 components resolving to its two real tables |
| Full rebuild adding a **new custom** table | Pinned (3/3) |
| Full rebuild adding a table that **already existed** — the reporter's step 2 | Pinned (4/4); `RetrieveAppComponents` answers **200** |
| The reporter's exact command `--only views,forms,app-shell --apply` | **Refused by the CLI**, and has been since the plugin's first commit — on `--apply` only a full build or exactly `--stage data` is allowed (design §14) |

I *did* reproduce a miss once, driving the SDK by hand with a fresh workspace that had never resolved the added table. A real build always resolves every spec entity in the data-model phase, so that state is not reachable through the plugin. I'm recording it rather than hiding it: it is the one case that behaved as reported, and it's why I am **not** claiming the engine is unconditionally correct.

On the accumulating `AppModuleComponent type="1" schemaName="entity"` rows: the plugin never calls `AddAppComponents` itself — only the SDK does, on create, and it pins by OData **reference**. So our code cannot produce those. They are most consistent with the manual Web API/designer reconciliation attempts described in the report.

## What is real, and is fixed

**`verify` could not tell the difference.** It confirms the table *exists* and that the sitemap *names* it — never that the app module *contains* it. So on an app in the reported state it printed **PASS**. That is exactly why the divergence was invisible, and the report is right to call it out.

`verify` now resolves the app's type-1 components to logical names and:

- **fails, naming the table**, when a sitemap-visible table is not a member;
- **fails closed** when the component list cannot be read — "could not look" must never read as "fine", which is the same shape as the bug;
- reports leftover **`entity` placeholder** components (the 6612527 corruption) even when every declared table is present;
- requires only **sitemap-visible** tables — a supporting table with no subarea is deliberately not pinned and must not fail a healthy app.

### Live proof

On a healthy 3-table app: `verify PASS (9/9)`, including the three new checks. Then I removed **one** table's component while leaving its subarea in place — the exact reported state:

```
✓ subarea: Gammas
✗ app-table-component: pp699_gamma
✗ verify FAIL — 1 missing (8/9 present)      exit 1
```

Before this change that same app reported PASS. Probe torn down (6 deleted, 0 failed).

## Also corrected

A stale comment in `download-model-app.js` asserted type-1 rows are unusable because "every row carries the same objectid — the `entity` metadata table". That was measured on an app corrupted by 6612527. **Re-measured on a healthy app the rows carry real table MetadataIds** — which is exactly what this new check relies on — so the claim could not be left standing. The view/chart/form derivation there is unchanged; it recovers lookup/sub-grid-only tables (6603388), which is its purpose.

## Verification

6 new tests; **5 mutations, all caught** (one was first written as a no-op and rewritten). Full suite **2454/2454**.

⚠️ As with the earlier stacked PR: every workflow is gated on `pull_request: branches: [main]`, so the test suites do **not** run on a PR targeting another branch. A `CLEAN` state here means nothing ran. The suite was reproduced locally and will run in CI once #563 merges and this retargets.


akshaymaloo:
Folded into #563 so there is a single PR to review. The commit (`c823a616`) is now on `users/akmaloo/dry-run-and-self-parent`, which is this PR's base — so this branch's head and #563's head are the same commit and nothing is lost.

Consolidating also fixes the CI gap noted in the description: #563 targets `main`, so the path-filtered `model-apps-script-tests` and `model-apps-evals` workflows actually run, which they never did while this PR was stacked.

The work-item references were also dropped from the code comments and test names in favour of plain descriptions.

PR #570, opened 2026-09-15
fix(power-automate): verify state after publish conflict

### LINKED ISSUE #407: [Bug] publish_flow reports success/alreadyStarted while flow remains Stopped
### Plugin

power-automate

### Plugin Version

2.5.0

### Skill / Command

build-flow / publish_flow

### Bug Description

`publish_flow` can return a success response claiming the flow is already enabled, while the flow's actual state (per `list_flows` / `get_flow`) is still `Stopped`. Callers cannot trust the tool's own response to confirm the flow is running.

### Steps to Reproduce

1. Create a new cloud flow that requires a connection which doesn't yet exist/isn't fully configured in the target environment (e.g. a connector that needs manual one-click confirmation, such as RSS).
2. Call `publish_flow` on that flow.
3. Call `list_flows` (or `get_flow`) immediately after to check the flow's `state`.

### Expected Behavior

`publish_flow` should either successfully start the flow and have `list_flows`/`get_flow` reflect `state: "Started"`, or return an error/false explaining why it could not start (e.g. missing connection).

### Actual Behavior

`publish_flow` returned:
```json
{"success": true, "alreadyStarted": true, "message": "Flow is already enabled."}
```
but a subsequent `list_flows` call showed the same flow's `state` as `"Stopped"`.

```json
{"success": true, "alreadyStarted": true, "message": "Flow is already enabled."}
```
followed by `list_flows` showing `"state": "Stopped"` for the same flow ID.

### Environment

- OS: Microsoft Windows 11 Pro Insider Preview (10.0.26220 Build 26220)
- Claude Code version: 2.1.220
- PAC CLI version: not installed / not on PATH

---
🤖 *This issue was created using the `/report-issue` skill.*

aakashkavuru101:
## Summary
- handle `CannotStartUnpublishedSolutionFlow` before the generic HTTP 409 path
- verify the live flow state before treating a 409 as already started
- preserve the original API error when the flow is stopped or its state cannot be confirmed

Fixes #407.

## Validation
- `node --check plugins/power-automate/server/mcp.mjs`
- `node scripts/validate-legacy-compatibility.js`
- `node scripts/validate-keyword-case.js`
- `node scripts/validate-no-real-environments.js`


aakashkavuru101:
@microsoft-github-policy-service agree company="オールブルー株式会社"

PR #573, opened 2026-09-15
model-apps: reconstruct relationships[] on download, and stop a downloaded spec failing its own lint

### LINKED ISSUE #567: [model-apps][Bug] download does not reconstruct relationships[], and does not report the omission
## What

`download-model-app.js` does not reconstruct `relationships[]`, and — unlike `forms[]`, `views[]`, `charts[]`, `businessRules[]` and `globalChoices[]` — **nothing reports the omission**. The "not round-tripped" warning lists five artifact classes; relationships is not one of them, so the spec looks complete.

## Why it matters

Live-measured on a one-table app with a **self-referencing** `OneToMany` (`pp_org` → `pp_org`, lookup `pp_ParentOrgId`):

- The build created the relationship, and sample data bound a two-level hierarchy through it (`Head Office` → `North Branch` → `North Sub`).
- The download emitted **no `relationships[]` key at all**.
- Rebuilding the downloaded spec into a fresh identity therefore produced a table with **no lookup and no hierarchy**, reporting success.

Until the fix in the same PR as #564, this was worse than a silent omission: the lookup's synthetic `<lookup>name` attribute *was* emitted as a real `Text` column, so the fresh rebuild created an invented `pp_parentorgidname` text field where the relationship used to be. That half is fixed — the shadow is now filtered — which makes the remaining gap a clean, honest absence rather than a wrong shape. But it is still an unreported absence.

## Scope

Two separable pieces, and the second is much cheaper than the first:

1. **Report it.** Add relationships to the `notRoundTrippedSummary` class list so a downloaded spec says the relationships are not carried. Note the existing comment in that function explains why `commands[]` is deliberately absent — there is no inventory to count. The same question needs answering for relationships: inventory them first, then add the class, so the report never emits a misleading "0 relationships".
2. **Reconstruct it.** `fetchEntityMetadata` already returns `relationships` and each attribute's `targets`, so a `OneToMany` between two downloaded tables looks reconstructible. Care is needed for: the relationship schema name vs the lookup schema name (they must differ — Dataverse rejects a collision), `ManyToMany` intersect tables, relationships pointing at tables outside the app, and the system-table bridge case.

## Evidence

Downloaded spec for the app above (abridged):

```jsonc
{
  "entities": [ { "schemaName": "pp_org", "columns": [ { "schemaName": "pp_tier", "type": "Choice", "options": ["Gold","Silver","Bronze"] } ] } ],
  // no relationships key
  "notRoundTripped": { "classes": [ { "kind": "forms" }, { "kind": "views" }, { "kind": "globalChoices" } ] }
}
```

Found while live-verifying the download round trip for #564; filed separately rather than folded in, because reconstructing relationships is a feature-sized change to the download path and not a regression from that work.

### LINKED ISSUE #572: [model-apps][Bug] download emits dashboard tiles the linter cannot resolve, so the downloaded spec fails its own structural lint
### Summary

`download-model-app.js` produces an `app-spec.json` that **fails its own `lint-app-spec.js` structural profile**. Downloading a real model-driven app that contains a dashboard and immediately linting the result yields 6 errors, so the download → lint → rebuild round-trip is broken for any app with a dashboard.

This is closely related to #567 (download does not reconstruct `relationships[]`); this issue is the dashboard-tile case.

### Repro

```bash
node scripts/download-model-app.js --env <envUrl> --app <appWithADashboard> --out ./out
node scripts/lint-app-spec.js --spec @./out/app-spec.json --profile structural
```

Observed (exit 1):

```
FAIL [profile: structural] — 6 error(s), 2 warning(s)
ERROR  lint: Dashboard 'Field Ops Overview' chart tile references unknown chart 'undefined'
ERROR  lint: Dashboard 'Field Ops Overview' chart tile references unknown view 'undefined'
ERROR  lint: Dashboard 'Field Ops Overview' list tile references unknown view 'undefined'
... (6 total)
```

### Root cause

Two coupled gaps.

**1. Tiles are emitted with GUID references, but the linter resolves by name.**

Download writes tiles shaped like this:

```jsonc
{
  "type": "chart",
  "name": "By Priority",
  "entity": "cfo_workorder",
  "viewId": "<guid>",
  "visualizationId": "<guid>"
}
```

The linter looks for `tile.chart` / `tile.view` (names). Those properties are absent, so it stringifies `undefined` and reports `references unknown chart 'undefined'`. The literal `'undefined'` in the message is itself a bug — an absent reference should not be rendered as if it were a name.

**2. There is nothing for a name to resolve against.**

In the same downloaded spec, `charts[]` and `views[]` are both empty (`0` entries) — they are intentionally not reconstructed and are recorded under `descriptionInventory` instead. So even if the tiles referenced charts/views by name, resolution would still fail.

### Expected

Any of the following would close the round-trip; the first is preferred:

- Reconstruct the dashboard's referenced `views[]`/`charts[]` and emit tiles that reference them by name.
- Or teach the linter to accept GUID-form tile references (`viewId`/`visualizationId`) and validate them against `descriptionInventory` rather than `charts[]`/`views[]`.
- Or, at minimum, have download report dashboards as not-round-trippable the way it already reports forms/views/charts/businessRules/globalChoices, so the omission is disclosed instead of producing an invalid spec.

Independently of which path is taken, the linter should never print `'undefined'` as a reference name; an absent reference deserves its own message.

### Notes

- **Not a regression.** Verified identical on the commit immediately before #563 and on `main` after it — same 6 errors, same 2 warnings. Pre-existing.
- Reproduced live against a real environment with a 3-table app containing one dashboard with 2 chart tiles and 2 list tiles.
- `preview-app.js` and `preview-form.js` both accept the same downloaded spec and exit 0, so only the lint path surfaces this.

akshaymaloo:
Fixes #567. Fixes #572.

Both are round-trip defects in `download-model-app.js` with a shared cause: the download emits a spec the rest of the toolchain does not fully agree with.

## #572 — a downloaded spec failed its own lint

A downloaded dashboard carries **ID-passthrough** tiles: the deployed view/chart ids rather than names, because those artifacts already exist and `views[]`/`charts[]` are deliberately not reconstructed. `validateAppSpec` (`app-spec.js`) and the build (`sdk-build.js`) have always accepted that form — `spec-lint.js` had never been taught it.

So `lint-app-spec.js` rejected a freshly downloaded spec with six errors reporting a chart literally named `'undefined'` (a missing key interpolated straight into the message).

The lint now understands id-passthrough tiles, still validates name-based ones, and reports a missing reference **as missing**.

## #567 — relationships[] was absent, and nothing said so

`relationships[]` was not reconstructed and — unlike forms, views, charts, business rules and global choices — was not in the not-round-tripped report either, so a downloaded spec *looked* complete while a rebuild into a fresh environment produced tables with no lookups and no hierarchy, reporting success.

1:N and N:N are now read from live metadata, keeping the lookup's deployed casing and label.

**Derived from the child side** (`ManyToOneRelationships`). That is how App Spec indexes a 1:N — `referencing` is the table that carries the lookup — and it is the only direction that also catches the documented *bridge to a standard table* pattern, where the parent is `systemuser`/`account` and so is never one of the downloaded entities.

**Raw metadata endpoints, not the SDK projection.** Live-measured, `fetchEntityMetadata().relationships` lacks all three facts this needs: no `IsCustomRelationship` (on a 3-table app, 20 of 22 entries per table were platform plumbing — `SyncErrors`, `AsyncOperations`, `MailboxTrackingFolders`…), no properly-cased lookup `SchemaName` (it lowercases `cfo_CustomerId` to `cfo_customerid`), and no `ManyToMany` entries.

Three cases App Spec **cannot** express are reported by name and reason rather than dropped silently:

| Case | Why |
|---|---|
| **Polymorphic lookup** | One column targeting several tables surfaces as several relationships sharing one lookup attribute; `relationships[]` declares exactly one `referenced` per lookup, so emitting both would declare the same lookup twice and fail the rebuild |
| **Parent is a custom table outside the app** | A rebuild target would not have it |
| **N:N with a partner outside the app** | Same |

A bridge to a *standard* table is kept — every org has one.

### Why these are not in `notRoundTrippedSummary`

That report's value is that its claims are true, and both of its standing sentences would become false for a class that is now *mostly* reconstructed: "this download does not reconstruct …" and "every one is listed under `descriptionInventory`". Skipped relationships get their own accurate warning instead.

An explicit `schemaName` is emitted **only** when the deployed name diverges from the generated default *and* satisfies the publisher-prefix rule the lint enforces. Omitting a divergent name would make a rebuild into the same environment create a second relationship beside the existing one; emitting a foreign-prefix name would hand back a spec that fails its own lint — the very defect #572 is about.

## Live verification

Same app, same environment, same commands, before and after:

| | pre-fix (`origin/main`) | this branch |
|---|---|---|
| `relationships[]` | **absent** | **2 reconstructed**, `cfo_CustomerId` / `cfo_TechnicianId` with labels |
| lint `structural` | exit 1, 6 errors | **exit 0, 0 errors** |
| lint `design` / `plan` / `deploy` | exit 1, 6 errors | **exit 0, 0 errors** |

The polymorphic lookup (`cfo_billto` → `account`+`contact`) was correctly detected and reported on real data. A dry-run rebuild plans both relationships. Downloaded **columns are byte-identical** before and after (19 = 19), so nothing else moved.

## Testing

Every guard in the new code is mutation-tested — **12 of 12 mutants killed, 0 survivors**. The sweep also caught one of my own new tests passing for the wrong reason (a shared assertion satisfied by the N:N read while it was meant to cover the 1:N read), and a real defect in my first draft: a failed 1:N read `continue`d past the N:N read, which would have left an N:N both unreconstructed and unmentioned.

| Gate | Result |
|---|---|
| Plugin suite | **2510 / 2510** (+16) |
| CI-exact evals | **176 / 176** |
| Layer 1 / Layer 2 / app-builder runners | **13/13 · 13/13 · 5/5** |
| All four repository validators | clean |

## Noted, not fixed here

A **polymorphic** lookup's shadow columns (`cfo_billtoname`, `cfo_billtoyominame`) are still emitted as real Text columns. Verified **pre-existing** — byte-identical before and after this change — and fixing it means threading relationship knowledge into `entityFromMetadata`, which is a different change. Filed separately rather than widened into this PR.


akshaymaloo:
Superseded by #579, which now carries this work plus the form-topology fix (#575) and the polymorphic-shadow-column fix (#574) as a single merge point.

Commit `f134e7da` from this branch is already contained in #579's history (verified with `git merge-base --is-ancestor`), so nothing here is lost — closing to avoid merging the same commit twice.

PR #598, opened 2026-09-18
feat(mobile-apps): diagnose startup failures before runtime

kanu-shubham:
## Why

`debug-app` currently expects a live Metro session and a loaded native app. An incomplete installation, unresolved configuration export, failed pre-start script, or QR-opening problem can prevent that state entirely. The resulting advice to start the app and rerun the debugger does not diagnose the failure that prevented startup.

This independent, main-based change adds a bounded setup path while preserving the existing runtime monitor. It also makes the debugging skill easier to discover from template-generated projects.

This branch does not include the orchestration changes from #596. Both touch some of the same skill documents, so those edits should be reconciled if one PR lands before the other.

## What changes

- Add `/debug-app startup` and setup dispatch for install/startup/device-opening symptoms or missing live sessions. Healthy runtime sessions keep the existing cursor-based monitor.
- Add a local read-only inspector for manifest/lockfile consistency, installed versions, Node engine requirements, and required package entry-point resolution. It never executes app/package code, fetches packages, writes project files, or authorizes a repair.
- Classify installation, compatibility, access/network, pre-start/Metro, device connectivity, native support, and app/package-defect evidence before choosing a fix.
- Offer same-lock npm restoration only with explicit approval. Keep manifest/lock hashes unchanged, handle required lifecycle scripts explicitly, and stop on inconsistent locks or incompatible runtimes rather than guessing.
- Separately approve bounded Metro start/restart, preserve process ownership and `predev`, and require verification of the original command or device symptom. Successful installation or quiet logs alone cannot mean the app is fixed.
- Route an observed `create-mobile-app` launch failure through the same startup-only workflow, forwarding the project root, sanitized evidence, and prior attempts. Successful creation does not start indefinite runtime monitoring.
- Bundle concise app-local agent guidance, with portable Copilot/Claude entry points and explicit plugin-install suggestions when skills are unavailable. Preserve existing customer instruction files.
- Recommend trying `debug-app` for app problems before `report-issue`, without blocking explicit reports or cases where plugin loading itself is broken.

## Deliberate boundaries

- **Restoration is not an upgrade.** Dependency/Node upgrades remain outside `debug-app`; no automatic invocation of an unfinished upgrade workflow.
- No lockfile deletion, blind upgrades, guessed peer-dependency bypasses, host/MSAL source patches, automatic native rebuilds, or automatic plugin installation.
- No new always-on hooks, terminal watcher, browser runtime testing, or direct Metro HTTP probes. An unshared terminal failure cannot wake an idle agent.
- `status` remains read-only; `--no-fix` cannot mutate or control processes. The setup path is bounded and returns pending/blocked when device verification is unavailable.

## Validation

- **102 targeted tests passed**: local startup-inspection fixtures, startup approval/return contracts, template guidance and preparation preservation/rollback, existing creation/planning/template contracts, Metro collector behavior, and diagnostic redaction.
- Cases include absent installation/configuration, mismatched or malformed lockfiles, npm optional-dependency overrides, missing exports, ancestor/symlink resolution, explicit project selection, no-fix/status behavior, and pending device verification.
- Optional-lockfile follow-up: all **24 targeted inspector/startup tests passed**, including new cases for absent optional package entries versus present malformed/linked entries and missing required packages. Installed-status reporting remains independent.
- `node --check` passed for the inspection and preparation scripts.
- Skill-description validation, legacy metadata compatibility, and `git diff --check` passed.
- CI passed for `76d3efee`: mobile script tests on Windows/macOS/Linux with Node 20/22, template preparation/type-checking, and repository metadata/keyword checks.

Tests use local synthetic projects and instruction contracts, not a live tenant or native device. No dependency-restoration or Metro-restart operation was performed against a customer app while authoring this PR.


plugins/mobile-apps/scripts/inspect-startup.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+    for (const name of new Set([...Object.keys(declared), ...Object.keys(locked)])) {
+      if (declared[name] !== locked[name]) mismatches.push({ section, name });
+    }
+  }
+  for (const dependency of dependencies) {
+    const entry = data.packages[`node_modules/${dependency.name}`];
+    if (!entry || typeof entry.version !== 'string' || entry.link) {
+      mismatches.push({ section: 'packages', name: dependency.name });
+    }
```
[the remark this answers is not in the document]
kanu-shubham: Fixed in 76d3efee. Missing package metadata is skipped only when the dependency is declared optional. The guard uses Object.hasOwn rather than !entry so a present null/false/malformed or linked entry is still reported. Installed-status reporting is unchanged. Added regression coverage for optional absence, independently installed optional packages, malformed/link entries, missing required packages, and declaration drift. Both new tests reproduced the issue before the fix; all 24 targeted inspector/startup tests now pass locally.

plugins/mobile-apps/scripts/inspect-startup.js
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/env node
```
nikhil5642, line 1: Are we ensuring this will cover every possible startup case? Have we tested a scenario where the actual issue is not covered into our script but the copilot reliably catches it and works outside the script. 

Please test the scenario to ensure the current script doesn't make copilot too biased towards covered cases.

plugins/mobile-apps/scripts/inspect-startup.js
```diff
@@ -0,0 +1,226 @@
...
+    for (let index = 0; index < argv.length; index += 1) {
+      const flag = argv[index];
+      if (!['--working-dir', '--entry-point'].includes(flag) || !argv[index + 1]
+          || argv[index + 1].startsWith('--')) throw new Error('Invalid inspection arguments');
+      if (flag === '--working-dir') {
+        if (projectRoot) throw new Error('Duplicate --working-dir');
+        projectRoot = argv[++index];
+      } else entries.push(argv[++index]);
+    }
+    if (!projectRoot) throw new Error('--working-dir is required');
```
nikhil5642, line 215: Have we tested these commands in windows machine as well? There the syntax is little different

plugins/mobile-apps/scripts/prepare-mobile-template.js
```diff
@@ -164,6 +170,55 @@ function copySharedFiles(projectRoot, samplesRoot) {
   return { copied, preserved };
 }
 
+function assertPreparationPath(projectRoot, relativePath) {
+  // lstat each component: existsSync follows links and misses dangling ones.
+  // In particular, a customer-owned `.github` must never redirect guidance writes.
+  const components = relativePath.split('/');
```
nikhil5642, line 178: this will break in windows machine. 

Path is mac is "home/abc"
Path in windows is "C:\abc"

nikhil5642 (approved):
Approved with minor comments

plugins/mobile-apps/scripts/prepare-mobile-template.js
kanu-shubham: The split here is applied to repository-relative constants such as .github/copilot-instructions.md, not to the native projectRoot. The native absolute root remains intact and is combined using Node path.join, which handles Windows drive paths. I kept that behavior and clarified the distinction in 401e4f0f. Added an end-to-end preparation CLI regression that creates the nested guidance files under a native absolute root containing spaces, an apostrophe, and brackets, and verifies the copied bytes and forward-slash writer metadata. It ran and passed on the actual Windows Node 20 and 22 CI jobs, alongside the existing preservation/rollback and junction tests. Evidence: https://github.com/microsoft/power-platform-skills/actions/runs/35583181149/job/106280434558 (test: preparation CLI copies POSIX-relative guidance beneath a native root with spaces).

plugins/mobile-apps/scripts/inspect-startup.js
kanu-shubham: Yes. The script suite already ran on Windows with Node 20 and 22; 401e4f0f now adds a direct Windows PowerShell invocation test as well as CLI tests using native absolute roots with spaces, apostrophes, and brackets. I checked the new Windows job logs: all three cases executed and passed (not skipped). Each Windows job reports 326 passed, 0 failed; the skipped test is the separately opt-in live Dataverse contract. The workflow now also shows distinct Bash/Git Bash and PowerShell command forms: no Bash line continuation or || in Windows PowerShell 5.1, Set-Location -LiteralPath for the app root, and npm.cmd for lifecycle commands. This validates shell/path handling; it does not claim a real dependency restoration or native-device launch on Windows. Windows Node 20: https://github.com/microsoft/power-platform-skills/actions/runs/35583181149/job/106280434661 ; Windows Node 22: https://github.com/microsoft/power-platform-skills/actions/runs/35583181149/job/106280434558

plugins/mobile-apps/scripts/inspect-startup.js
kanu-shubham: No: the script is deliberately bounded, not coverage of every startup failure. In 401e4f0f the result now explicitly says validationScope: static-metadata-only and startupVerified: false. The workflow says an observed failure takes precedence over clean metadata and provides a scoped evidence-led investigation path, with no guessed reinstall or bypass of approval/no-fix/time limits.

I tested the scenario you requested, beyond a prose assertion: a synthetic project had a matching manifest/lock/install and a resolved package export, but its app-owned pre-start validator failed. Using the source debug-app workflow and the reported failure, diagnosis inspected the validator and launch configuration, found its reference to a nonexistent app/start.tsx, and changed only launch-config.json to the existing app/index.tsx. Rerunning the approved failed pre-start command returned exit 0 instead of exit 1. I independently compared the resulting files against the original fixture: manifest, lockfile, installed package, and every other file were unchanged. The inspector report was also unchanged, demonstrating that actual command evidence—not a green metadata report—proved this repair.

The returned outcome correctly remained verification-pending for full npm run dev / Metro / native-device behavior, which was not exercised. There is also a permanent automated regression for this case. This is a bounded behavioral smoke check, not a claim of exhaustive case coverage or guaranteed agent reliability; insufficient evidence must still return blocked with a concrete next step.

